PR #7, opened 2015-01-25
Slowdown a spec even more to avoid flapping tests

splattael:
Let's see...


splattael:
I don't think slowing down helps here. I have to investigate more...

solnic:
the build was passing consistently for quite a while...maybe the problem is gone...can we close this for now? :)

splattael:
Ups, I've totally forgot to close this one.

This flapping spec has already been fixed in f9bf56a78f184993b725c645a84dfb998082196a

Thanks for the heads-up!

solnic:
no worries :) thanks for taking care of this! :heart:

PR #8, opened 2015-01-29
#130 migrations

gotar:
Basic migration interface for ROM based on Sequel. 

few basic rake tasks already implemented in rom-sql:

```bash
$ bundle exec rake -T
rake db:clean                           # Perform migration down (erase all data)
rake db:create_migration[name,version]  # Create a migration (parameters: NAME, VERSION)
rake db:migrate[version]                # Migrate the database (options [version_number])]
rake db:reset                           # Perform migration reset (full erase and migration up)
```

To make it working you need to add rom-sql to your local Rakefile, and setup connection there, for example:

```ruby
# Rakefile
require 'rom-sql'
require_relative 'apps/api/db/db'

namespace :db do
  task :load_setup do
    ROM::SQL::Migration.connection = DB.setup_connection_to_db
    ROM::SQL::Migration.path = 'apps/api/db/migrations' # default path is: db/migrate. You can skip those line if its fit you
  end
end
```

and thats it ;]

TODO:
* nice Errors
* create_db, drop_db, ... and other tasks you know from AR and maybe need

### DIFF
--- a/lib/rom-sql.rb
+++ b/lib/rom-sql.rb
@@ -1,2 +1,2 @@
 require 'rom/sql'
-require 'task'
+require 'rom/rake_task'
--- a/lib/rom-sql.rb
+++ b/lib/rom-sql.rb
@@ -1,2 +1,2 @@
 require 'rom/sql'
-require 'rom/rake_task'
+require 'rom/sql/rake_task'
--- a/lib/rom-sql.rb
+++ b/lib/rom-sql.rb
@@ -1,1 +1,2 @@
 require 'rom/sql'
+require 'task'
--- a/lib/rom/sql/migration.rb
+++ b/lib/rom/sql/migration.rb
@@ -5,18 +5,20 @@ module ROM
 
       DEFAULT_PATH = 'db/migrate'
 
-      attr_accessor :path, :connection
+      class << self
+        attr_accessor :path, :connection
 
-      def path
-        @path || DEFAULT_PATH
-      end
+        def path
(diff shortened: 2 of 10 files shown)


lib/rom-sql.rb
```diff
@@ -1 +1,2 @@
 require 'rom/sql'
+require 'task'
```
solnic: This should be moved to `rom/sql/rake_task`

lib/rom/sql/migration.rb
```diff
@@ -0,0 +1,25 @@
+module ROM
+  module SQL
+    class Migration
+      ::Sequel.extension :migration
+
+      DEFAULT_PATH = 'db/migrate'
+
+      class << self
+        attr_accessor :path, :connection
```
solnic: since we override `path` then it could just be `attr_writer :path` I suppose

solnic:
Thanks :) That's a great start. I tried it locally, worked just fine. I need to think how to handle connection, in rom-rails we could do things automatically though so it's not a big issue. Otherwise having a convention that you gotta implement `load_setup` task would be good-enough too.

Apart from that - I had a couple of comments in the diffs.

lib/rom/sql/migration.rb
```diff
@@ -0,0 +1,25 @@
+module ROM
+  module SQL
+    class Migration
+      ::Sequel.extension :migration
```
elskwid: That is sweet!

solnic:
Merged thanks!

PR #10, opened 2015-02-02
#116 - Do not call db to read data after insert.

gotar:
Do not call DB again after data was inserted to DB, use returning function from Sequel to get them for free.

### DIFF
--- a/spec/unit/commands_spec.rb
+++ b/spec/unit/commands_spec.rb
@@ -1,0 +1,37 @@
+require 'spec_helper'
+
+describe ROM::SQL::Commands do
+  include_context 'users and tasks'
+
+  subject(:users) { rom.commands.users }
+
+  before do
+    setup.relation(:users)
+
+    setup.commands(:users) do
+      define(:create) do
+        result :one
+      end
+    end
+  end
+
+  context 'postgres' do
+    it 'create instance of PostgreSQL create command' do
+      expect(ROM::SQL::Commands::Postgres::Create)
+        .to receive(:new).and_call_original
+
+      users.try { create(id: 1, name: 'Foo') }
+    end
+  end
+
+  context 'others' do
+    it 'create instance of default create command' do
+      type = double(database_type: :foo)
+      allow_any_instance_of(ROM::Relation).to receive(:db).and_return(type)
+
+      expect(ROM::SQL::Commands::Create).to receive(:new).and_call_original
+
+      users.try { create(id: 1, name: 'Foo') }
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


lib/rom/sql/commands_ext/postgresql.rb
```diff
@@ -0,0 +1,19 @@
...
+  module SQL
+    module Commands
+      module PostgreSQL
+        class Create < ROM::Commands::Create
+          def execute(tuples)
+            Array([tuples]).flatten.map do |tuple|
+              attributes = input[tuple]
+              validator.call(attributes)
+              relation.returning(*attributes.keys).insert(attributes.to_h).first
+            end
```
solnic: We should have something like `Create#insert` that can be overridden and simply implement it here as:

``` ruby
def insert(attributes)
  relation.returning(*attribute.keys).insert(attributes.to_h).first
end
```

With the default implementation that has the original behavior of inserting then fetching the row again using pk.

WDYT?

lib/rom/sql/commands_ext/postgresql.rb
```diff
@@ -0,0 +1,19 @@
+module ROM
+  module SQL
+    module Commands
+      module PostgreSQL
```
solnic: Would be better to use "Postgres" because that's how it's called in Sequel, that's the database adapter type in config files and it doesn't have "SQL" acronym inside which can cause trouble due to naming conventions etc.

lib/rom/sql/commands_ext/postgresql.rb
gotar: Yeap why not i will add it

spec/unit/commands_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    it 'create instance of default create command' do
+      type = double(database_type: :foo)
+      allow_any_instance_of(ROM::Relation).to receive(:db).and_return(type)
+
+      expect(ROM::SQL::Commands::Create).to receive(:new).and_call_original
+
+      users.try { create(id: 1, name: 'Foo') }
+    end
+  end
+end
```
solnic: Could we have that spec w/o mocking? Just check that correct instance of postgres subclass is present and that it works.

gotar: in that way we will never test default behaviour because all tests use postgres behind the scene. I think one test to check default behaviour should be done, but maybe I'm not right? Do not see smt?

lib/rom/sql/commands_ext/postgresql.rb
gotar: yeap will change it

spec/unit/commands_spec.rb
solnic: Oh right of course you're right! We need tests for default behavior of course...dooh :)

PR #11, opened 2015-02-04
Dirty attributes update

solnic:
This adds a way to pass in an original tuple to update command and it will skip the update if nothing was changed:

``` ruby
users.try { update(:by_id, 1).change(name: 'Jane').to(name: 'Jane Doe') }
```

As a bonus it adds pg-specific update which uses `returning` :)


solnic:
I'm gonna merge this in since we use it in our app. Command DSL added in this PR is a subject to change and will be discussed further in rom-rb/rom#129

gotar:
nice, plus postgres specify update code ;]

PR #12, opened 2015-02-07
Basic transaction interface for commands

gotar:
Basic transaction interface on top of commands. All options for Sequel transactions works: http://sequel.jeremyevans.net/rdoc/files/doc/transactions_rdoc.html

Example:

```ruby
users = rom.commands.users

users.create.transaction do
  users.create.call(...)
end

users.create.transaction do
  users.create.call(...)
  tags.create.transaction do
    tars.craete.call(...)
  end
end

# will create nothing
users.create.transaction do
  users.create.call(...)
  raise Sequel::Rollback
end

# will try again if failure
users.create.transaction(rollback: :reraise) do
  users.create.call(...)
  raise Sequel::Rollback
end
```

### DIFF
--- a/lib/rom/sql/commands/transaction.rb
+++ b/lib/rom/sql/commands/transaction.rb
@@ -1,0 +1,17 @@
+require 'rom/commands/result'
+
+module ROM
+  module SQL
+    module Commands
+      module Transaction
+        ROM::SQL::Rollback = Class.new(Sequel::Rollback)
+
+        def transaction(options = {}, &block)
+          ROM::Commands::Result::Success.new(
+            relation.dataset.db.transaction(options, &block)
+          )
+        end
+      end
+    end
+  end
+end
--- a/spec/integration/commands/update_spec.rb
+++ b/spec/integration/commands/update_spec.rb
@@ -27,6 +27,25 @@ describe 'Commands / Update' do
     relation.insert(name: 'Piotr')
   end
 
+  context '#transaction' do
+    it 'update record if there was no errors' do
+      result = users.update.transaction do
+        users.update.by_id(piotr[:id]).set(peter)
+      end
+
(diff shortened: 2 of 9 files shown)


lib/rom/sql/commands/transaction.rb
```diff
@@ -0,0 +1,17 @@
+require 'rom/commands/result'
+
+module ROM
+  module SQL
+    module Commands
+      module Transaction
+        ROM::SQL::Rollback = Class.new(Sequel::Rollback)
+
+        def transaction(options = {}, &block)
+          ::ROM::Commands::Result::Success.new(
```
solnic: Can we just have `Result::Succes` here?

gotar: yeap will change it

spec/integration/commands/update_spec.rb
```diff
@@ -27,6 +27,25 @@ def by_name(name)
...
+      end
+
+      expect(result.value).to eq([{ id: 1, name: 'Peter' }])
+    end
+
+    it 'updates nothing if error was raised' do
+      users.update.transaction do
+        users.update.by_id(piotr[:id]).set(peter)
+        raise Sequel::Rollback
+      end
```
solnic: Should we introduce `ROM::SQL::Rollback` or use `throw/catch` and do `throw :rollback`?

gotar: eh I already add ROM::SQL::Rollback forget to change in test, will do it ;]

lib/rom/sql/commands/transaction.rb
gotar: I check it and not, path is not the same because there is ROM::SQL::Commands here not ROM::Commands i can only remove those :: before ROM

PR #13, opened 2015-02-08
Implement `group by` and friends

mcls:
Trying to get my head around how this would work with ROM.

I wrote an integration spec to show what kind of things I'm trying to do.

What do you think?

### DIFF
--- a/lib/rom/sql/relation.rb
+++ b/lib/rom/sql/relation.rb
@@ -106,6 +106,18 @@ module ROM
         __new__(dataset.__send__(__method__, *args, &block))
       end
 
+      # Append specific columns to select clause
+      #
+      # @example
+      #   users.select(:id, :name).select_append(:email)
+      #
+      # @return [Relation]
+      #
+      # @api public
+      def select_append(*args, &block)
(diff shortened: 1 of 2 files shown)


solnic:
:+1: could you add yard docs and push this to master? :)

lib/rom/sql/relation.rb
```diff
@@ -39,6 +39,26 @@ def project(*names)
...
+
+      # @api public
+      def select_append(*args, &block)
+        __new__(dataset.__send__(__method__, *args, &block))
+      end
+
+      # @api public
+      def select_group(*args, &block)
+        __new__(dataset.__send__(__method__, *args, &block))
+      end
```
gotar: why not to use some alias method? all those methods do the same for me

mcls: They're being defined separately so they can be documented. But I guess the `__new__(dataset.__send__(__method__, *args, &block))` lines "DRY'ed" a bit.

solnic: we could dry it up but I'm not sure if it's worth it, those methods will not change at all. As @mcls pointed out it's just to define the interface explicitly and document it.

PR #14, opened 2015-02-17
Use internal ROM interface for inflector

mjtko:
With changes merged from rom-rb/rom#158, `rom` now provides an internal API to access an inflector implementation.  This change updates `rom-sql` to use that API and removes any dependency on the `inflecto` gem.


solnic:
thanks!

PR #15, opened 2015-02-25
Fallback pagination

gotar:


solnic:
I don't think it's a good idea actually. This is error-prone and may lead to confusing debugging sessions when you changed name of param and yet rom defaults it for you and you wonder why things don't work.

IMO every more complex query that relies on user input where you can use default values should be handled by some query object. Typically I use virtus for that.

gotar:
Then you think everybody should use everywhere they use pagination smt like

```
PaginationValidation.call(PaginationParams.new(params))
```

solnic:
Yeah as soon as the relation needs parameterized input I go for a query object. In fact I'm planning to build a rom add-on with such functionality.

gotar:
The only disadvantage is to manually write and add it every time, but maybe its better solution, ok I will close those PR

solnic:
This would make rom-sql responsible for handling user input in web-apps, it's a similar mistake as input sanitization happening on an ORM-level.

PR #18, opened 2015-04-02
Let command.try{} catch Sequel::DatabaseError

kwando:


kwando:
See https://github.com/rom-rb/rom-sql/pull/19 instead

PR #19, opened 2015-04-02
The create command now rescues from Sequel::DatabaseError

kwando:


solnic:
Thanks. Could you add a test for it too? Should be simple, like you mentioned in another issue, just try to use invalid column names or something and see if it returns a `Result::Error` object with correct command error wrapped in its value.

kwando:
Added a test for this as requested.

kwando:
Shouldn't the Update and Delete command have some error handling too? I think you can have both SQL and constraint errors there..

solnic:
@kwando yep all commands should have proper support for db errors

kwando:
Can I assume a ruby version >= 2.0? ```Module#prepend``` would fit the bill nicely in this case.

solnic:
@kwando yes if it works on jruby and rbx

kwando:
Added error handling with ```#prepend```, this will not work with jruby 1.7.x in the default 1.9 mode =/

solnic:
Can we somehow make it work without prepend? We shouldn't break jruby compatibility in a patch version bump, even if it's 1.9 mode still.

solnic:
I just realized that we can simply have a `ErrorHandler` module that overrides `#call(*args)` and rescue errors there and then include it in all command classes.

solnic:
@kwando I wrapped this up and merged in your commits with a couple of tweaks so that we don't need `prepend`. thanks for help :)

kwando:
Ah, cool! Just did the changes you requested now and was surprised my pull request was gone =P

solnic:
@kwando oh sorry about that. I had very little time and needed to push the release :/

kwando:
No worries, I'm just glad it is done =)

Skickat från min iPhone

PR #20, opened 2015-04-03
[WIP] Add Migrator repository plugin

solnic:
This makes it possible to inject a migrator to repository which exposes
migration DSL and interface required for running migrations.

This means that the only thing we need to do to use migrations is to
simply setup a repository which removes the `load_setup` task
requirement.


gotar:
+1 really nice

PR #21, opened 2015-04-04
Fix Paginator::total_pages behavior

larribas:
Hello, 

I detected what I believe is a behavioral error at ``Paginator::total_pages``. Correct me if I'm wrong.

### DIFF
--- a/lib/rom/sql/plugin/pagination.rb
+++ b/lib/rom/sql/plugin/pagination.rb
@@ -31,7 +31,7 @@ module ROM
           end
 
           def total_pages
-            (total / per_page) + 1
+            (total / per_page.to_f).ceil
           end
 
           def at(dataset, current_page, per_page = self.per_page)
--- a/spec/unit/plugin/pagination_spec.rb
+++ b/spec/unit/plugin/pagination_spec.rb
@@ -53,6 +53,18 @@ describe 'Plugin / Pagination' do
     end
   end
 
+  describe '#total_pages' do
+    it 'returns a single page when elements are a perfect fit' do
+      users = rom.relation(:users).page(1).per_page(3)
+      expect(users.pager.total_pages).to eql(3)
+    end
+
+    it 'returns the exact number of pages to accommodate all elements' do      
+      users = rom.relation(:users).per_page(9)
+      expect(users.pager.total_pages).to eql(1)
+    end
+  end
+
   describe '#pager' do
     it 'returns a pager with pagination meta-info' do
       users = rom.relation(:users).page(1)
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@
 [![Test Coverage](https://codeclimate.com/github/rom-rb/rom-sql/badges/coverage.svg)][codeclimate]
 [![Inline docs](http://inch-ci.org/github/rom-rb/rom-sql.svg?branch=master)][inchpages]
 
-RDBMS suport for [Ruby Object Mapper](https://github.com/rom-rb/rom).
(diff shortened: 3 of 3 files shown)


gotar:
Looks good, but please add corresponding tests to: spec/unit/plugin/pagination_spec.rb If you don't know how we can help you

larribas:
@gotar Done

spec/unit/plugin/pagination_spec.rb
```diff
@@ -53,6 +53,16 @@
...
   end
 
+  describe '#total_pages' do
+    it 'returns the exact number of pages to accommodate all elements' do
+      users = rom.relation(:users).page(1).per_page(3)
+      expect(users.pager.total_pages).to eql(3)
+      
+      users = rom.relation(:users).per_page(9)
+      expect(users.pager.total_pages).to eql(1)
+    end
```
gotar: this should be split to two different tests.

.gitignore
```diff
@@ -13,3 +13,4 @@
 *.a
 mkmf.log
 log/*.log
+.ruby-version
```
gotar: Its not sth you should keep it project .gitignore file, but your global one: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore because its related to your local setup

lib/rom/sql/plugin/pagination.rb
```diff
@@ -31,7 +31,9 @@ def total
           end
 
           def total_pages
-            (total / per_page) + 1
+            total_pages = total / per_page
+            total_pages += 1 unless total % per_page == 0
+            total_pages
```
gotar: `(r = total/per_page) == 1 ? 1 : r` but not sure what I like more @solnic ?

larribas: That one won't work because we are looking for whole ratios (with no decimal positions). Thus the two expectations in the spec ;)

I thought of this oneliner
`(total/per_page) + (total % per_page == 0 ? 0 : 1)`
But seemed too awkward

spec/unit/plugin/pagination_spec.rb
larribas: If you want, I can separate them, but I think they should be together. They are testing the same property: that `total_pages` only adds +1 when the number of elements divided by the page size is a whole number (no decimals)

.gitignore
larribas: Noted. Thanks

spec/unit/plugin/pagination_spec.rb
gotar: they test the same but single it block should have only one expect block. You have describe or (context) to group them but here you should have two tests. One for every scenario you test.

lib/rom/sql/plugin/pagination.rb
gotar: ok leave the code as it is

hecrj: What about...

```ruby
(total / per_page.to_f).ceil
```

gotar: +1

larribas: Nice one!

For further reference, Sequel's [pagination module](https://github.com/jeremyevans/sequel/blob/master/lib/sequel/extensions/pagination.rb) considers that the dataset would have 1 page when there are no elements at all.

```ruby
@page_count = (record_count / page_size.to_f).ceil
@page_count = 1 if @page_count == 0
```

I haven't. I'll leave to you such decision

solnic:
Thanks!

PR #24, opened 2015-05-04
Add fixes to README example

nepalez:
Hi, Piotr!
After playing with the README example, I've added small fixes to make it exactly reproducible


solnic:
Thank you (8

PR #25, opened 2015-05-09
fixed bug in ROM::SQL::Relation::Associations#graph_join

nepalez:
Used the name of the dataset (db table) instead of the relation's name,
that can differ from it.

See [rom#218](https://github.com/rom-rb/rom/issues/218)


solnic:
Thanks! :)

PR #26, opened 2015-05-25
Wrap errors  when calling command with []-syntax

kwando:
Fixes rom-rb/rom#236

Note: I'm only testing that [] works for one case, this should probably be changed so that we run all error re-raising tests with both ```[]``` and ```call```.


solnic:
Thanks this is good enough.

PR #29, opened 2015-06-06
Add helper method to set table name on a relation

vrish88:
This is the best implementation I could figure out. I tried to specify the table name using the model syntax  (eg `class Post < Sequel::Model(:my_posts); end`) that Sequel provides but I couldn't figure out how to get a handle on it [at inheritance](https://github.com/rom-rb/rom-sql/blob/b231c11c1cc0b520d7b15bc1b89761fd3b0b9ea5/lib/rom/sql/relation/class_methods.rb#L17).

Is there a better way to do this? It feels a little janky having to set the table name twice in the finalize method.

### DIFF
--- a/lib/rom/sql/relation/class_methods.rb
+++ b/lib/rom/sql/relation/class_methods.rb
@@ -93,12 +93,29 @@ module ROM
           associations << [__method__, name, new_options]
         end
 
+        # Specify the table for the model
+        #
+        # @example
+        #   class Tasks < ROM::Relation[:sql]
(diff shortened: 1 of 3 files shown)


solnic:
In ROM table name == `dataset`. Is that not working for you?

ps. I completely missed that PR, sorry about that

solnic:
This is fairly outdated so it's been closed. I should mention that using `Sequel::Model` will be removed prior 1.0.

PR #31, opened 2015-07-20
Multi insert in create command

draxxxeus:
This will enable the use of multi_insert whenever the create command receives an array of tuples


gotar:
+1, please squash

AMHOL:
:+1:

PR #32, opened 2015-07-23
Relation name and join key fixes for :many_to_one

jamesmoriarty:
@solnic Thanks for all your hard work. I'm really enjoying using ROM. I've been using it to make a legacy database a little more consistent and I think I may have found a couple of bugs.

I would love it if you could confirm this is a bug or maybe I'm just _not_ using ROM incorrectly?

1. Can't change `many_to_one` relation name.
2. `many_to_one` uses the wrong tables primary key name.

### DIFF
--- a/spec/unit/many_to_one_spec.rb
+++ b/spec/unit/many_to_one_spec.rb
@@ -43,4 +43,18 @@ describe 'Defining many-to-one association' do
       [{ id: 1, title: 'Finish ROM', user: { name: 'Piotr' } }]
     )
   end
+
+  it "joins on specified key" do
+    setup.relation(:task_tags) do
+      many_to_one :tags, key: :tag_id
+
+      def with_tags
+        association_left_join(:tags)
+      end
+    end
+
+    expect(rom.relation(:task_tags).with_tags.to_a).to eq(
+      [{ tag_id: 1, task_id: 1, id: 1, name: "important" }]
+    )
+  end
 end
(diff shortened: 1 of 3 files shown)


jamesmoriarty:
PS: I'll cleanup the test.

spec/unit/many_to_one_spec.rb
```diff
@@ -43,4 +43,34 @@ def with_user
...
+      end
+    end
+
+    conn[:order].insert id: 1
+    conn[:purchase].insert sku: 2, orderid: 1
+
+    expect(rom.relation(:purchase).with_order.to_a).to eq(
+      [{ sku: 2, orderid: 1, id: 1 }]
+    )
+  end
```
solnic: There's a shared db setup context where we clear test tables

solnic:
Hey James, this is a good change thank you for spotting this *and* fixing :) I'll wait for you to clean up the tests like you said and will merge it in and release new version.

BTW I stopped using those macros as I discovered I'm perfectly happy with the lower level `join` / `outter_join` interface. Renaming and qualifying columns is as easy as `select(:id, :name).qualified.prefix(:bazinga)` so I really no longer feel the need for those macros. Not saying they should be removed but I figured I should mention that :)

jamesmoriarty:
@solnic thanks for getting back to me so quick. I'll try the lower level method tomorrow. 

I've cleaned up the test using the existing database shared example. Looks much better.

solnic:
Thanks, I just released 0.5.3 with your fix included :tada:

PR #34, opened 2015-07-31
bugfix: many_to_one pluralize default relation option.

jamesmoriarty:
@gotar @solnic restore default behavior but allows the ability to override relation option.

ref: https://github.com/rom-rb/rom-sql/commit/196d703239e6bad2ac26508ca847f7b32fa1896b

### DIFF
--- a/lib/rom/sql/relation/class_methods.rb
+++ b/lib/rom/sql/relation/class_methods.rb
@@ -65,7 +65,7 @@ module ROM
         #
         # @api public
         def many_to_many(name, options = {})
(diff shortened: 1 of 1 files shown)


gotar:
Hmm not sure we should make step back. Just add better error message, or some doc. 

The best will be to drop those macros I think and just define everything in relation method.

solnic:
We could add a check if that first arg matches any relation and if not assert presence of `:relation` option with a valid relation name.

This would fix the situation, right?

PR #36, opened 2015-08-21
Add `:extensions` option to Gateway

c0:
- Array of extensions is passed onto the connection.

Example Usage:
`ROM.setup(:sql, [‘postgres://localhost/db', extensions: [:pg_enum, :pg_array])`

@solnic how would you suggest testing and documented this?

### DIFF
--- a/spec/unit/gateway_spec.rb
+++ b/spec/unit/gateway_spec.rb
@@ -35,6 +35,15 @@ describe ROM::SQL::Gateway do
 
       expect(gateway.options).to eql(migrator: migrator)
     end
+
+    it 'allows extensions' do
+      extensions = [:pg_array, :pg_enum]
+      connection = Sequel.connect uri
+
+      expect(connection).to receive(:extension).with(:pg_array, :pg_enum)
+
+      ROM::SQL::Gateway.new(connection, extensions: extensions)
+    end
   end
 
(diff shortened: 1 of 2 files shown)


solnic:
I would mock it, you can pass in `connection` object directly to setup via `ROM.setup(:sql, [connection]`) so you can pass a spy and expect it to `have_received(:extension)` with expected arguments. It is not our interface but the risk that this interface will change in Sequel is very small. This would be way less overhead than adding a pg enum column to test table and testing the whole thing in integration style.

c0:
@solnic specs added

spec/unit/gateway_spec.rb
```diff
@@ -35,6 +35,15 @@
 
       expect(gateway.options).to eql(migrator: migrator)
     end
+
+    it 'allows extensions' do
+      extensions = %i(pg_array pg_enum)
```
solnic: jruby doesn't support `%i` so the build failed :)

c0: Updated. Waiting for CI :)

solnic:
Awesome, thanks :)

solnic:
Could you update changelog too? Just push to master directly if you have time, you've got commit access now

c0:
Done :D

PR #38, opened 2015-09-27
Add `union` method for relations

spscream:
I added ability to make a unions on relations


spscream:
I'm not sure if this is a right way to do this thing. But it works in my case.

solnic:
Thanks Alexander :)

PR #44, opened 2015-11-04
Allow to check for pending migrations

gotar:
Using internal Sequel mechanism allow to check if there are pending
migrations:

```
ROM.env.gateways[:default].pending_migrations?
```

### DIFF
--- a/lib/rom/sql/migration.rb
+++ b/lib/rom/sql/migration.rb
@@ -55,6 +55,13 @@ module ROM
         end
       end
 
+      # @see ROM::SQL::Migration.pending?
+      #
+      # @api public
+      def pending_migrations?
+        migrator.pending?
+      end
+
       # @see ROM::SQL.migration
       #
       # @api public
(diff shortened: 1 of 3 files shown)


lib/rom/sql/migration.rb
```diff
@@ -55,6 +55,13 @@ def self.included(klass)
         end
       end
 
+      # @see ROM::SQL.pending?
```
solnic: This should be `ROM::SQL::Migration.pending?` I believe :)

gotar: Eh yes, fixed

solnic:
Thanks!

solnic:
@gotar could you mention the addition in changelog too?

gotar:
yeap

nepalez:
It is interesting, how  this could be abstracted to `rom-migration`

gotar:
`pending_migrations?` is our public API, what is below its not important, then other adapters can implement it in own way.

nepalez:
I'm just slowwwwwly trying to implement abstract base migrator there. For now it is about ready to be used by cassandra, but with `rom-sql`... not sure. Maybe 90%

gotar:
I completly forgot we have external rom-migrator, sql do not use it at all, I can try to rewrite it in a way to use those external migration gem maybe in next week

nepalez:
It would be nice if you could give a feedback about that migrator

PR #58, opened 2016-04-11
Materialize relation on single insert, fixes #57

### LINKED ISSUE #57: Create#insert doesn't materialize returned relation
We gotta `to_a` on returned relation, otherwise weird stuff happens when a command with `:many` results is used as it returns a relation, in *some* cases it may crash. We probably didn't notice because sequel's dataset API has a lot of common interfaces with `Array`.

Duck-typing FTW and everything but in this case we do need to materialize the relation, because that's the contract.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32751848-create-insert-doesn-t-materialize-returned-relation?utm_campaign=plugin&utm_content=tracker%2F8353094&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F8353094&utm_medium=issues&utm_source=github).
</bountysource-plugin>

AMHOL:


solnic:
Could you add a test for this using sqlite? Here's a sample setup:

``` ruby
container = ROM.container(:sql, 'sqlite::memory') do |config|
  config.gateways[:default].connection.create_table :users do
    primary_key :id
    String :name, null: false
  end
end
```

We could have it in the shared examples, because it's already used in relation_spec.rb

AMHOL:
Will have to do that tonight

spec/integration/commands/create_spec.rb
```diff
@@ -278,4 +280,64 @@ def self.[](input)
...
+        end
+      end
+
+      context 'with multiple records' do
+        subject! do
+          container.command(:users).create.execute([
+            { name: 'Jane' },
+            { name: 'John' }
+          ])
+        end
```
solnic: could you get rid of that and simply assign result to a local var in the example block?

PR #61, opened 2016-04-27
Add support for the rest of associations

flash-gordon:


flash-gordon:
@solnic just copypasted your code ><

PR #63, opened 2016-05-21
Add basic schema inferrer

flash-gordon:


solnic:
This is a great start. I'm not entirely sure if we should use the dsl object to create attributes but we can live with that for now.

PR #66, opened 2016-06-08
Add upsert option for insert call. Only works for PostgreSQL >=9.5

gotar:
using Sequel native support

### DIFF
--- a/lib/rom/sql/commands/postgres.rb
+++ b/lib/rom/sql/commands/postgres.rb
@@ -5,13 +5,30 @@ module ROM
         class Upsert < SQL::Commands::Create
           adapter :sql
 
+          defines :constraint, :conflict_target, :update_statement, :update_where
+
+          option :constraint, reader: true, default: -> c { c.class.constraint }
+          option :conflict_target, reader: true, default: -> c { c.class.conflict_target }
+          option :update_statement, reader: true, default: -> c { c.class.update_statement }
+          option :update_where, reader: true, default: -> c { c.class.update_where }
+
           # @api private
           def execute(tuples)
             inserted_tuples = with_input_tuples(tuples) do |tuple|
-              upsert.call(input[tuple])
+              upsert(input[tuple], upsert_options)
             end
 
             inserted_tuples.flatten
+          end
+
+          # @api private
+          def upsert_options
+            @upsert_options ||= {
+              constraint: constraint,
+              target: conflict_target,
+              update_where: update_where,
+              update: update_statement
+            }
           end
         end
       end
--- a/lib/rom/sql/commands/postgres.rb
+++ b/lib/rom/sql/commands/postgres.rb
@@ -1,0 +1,20 @@
+module ROM
+  module SQL
+    module Commands
+      module Postgres
+        class Upsert < SQL::Commands::Create
+          adapter :sql
+
+          # @api private
+          def execute(tuples)
+            inserted_tuples = with_input_tuples(tuples) do |tuple|
+              upsert.call(input[tuple])
+            end
+
+            inserted_tuples.flatten
+          end
+        end
+      end
+    end
+  end
+end
--- a/spec/integration/commands/upsert_spec.rb
+++ b/spec/integration/commands/upsert_spec.rb
@@ -6,21 +6,78 @@ RSpec.describe 'Commands / Postgres / Upsert', adapter: :postgres do
   let(:tasks) { commands[:tasks] }
 
   before do
-    conf.commands(:tasks) do
-      define('Postgres::Upsert') do
-        register_as :create_or_update
-        result :one
-      end
-    end
+    conn[:users].insert id: 1, name: 'Jane'
+    conn[:users].insert id: 2, name: 'Joe'
+    conn[:users].insert id: 3, name: 'Jean'
   end
 
   describe '#call' do
-    let(:task) { { title: 'task 1' } }
+    let(:task) { { title: 'task 1', user_id: 1 } }
+    let(:excluded) { task.merge(user_id: 3) }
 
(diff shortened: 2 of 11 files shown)


gotar:
@solnic WDYT? I was thinking a lot, in which direction it should go and how to implement this, but this one win. But I'm open to any suggestions? Not sure you want sth like this?

solnic:
I updated this branch and added an explicit `Postgres::Upsert` command class. This will allow us to fine-tune `upsert` behavior per-command basis.

I think this needs a bit more work so that it's configurable. Personally I'd like to achieve a smart `update_or_delete` behavior. It seems like it's possible via `dataset.insert_conflict(custom_opts).insert(tuple)`.

@flash-gordon do you have any experience with this?

flash-gordon:
@solnic wdym by `update_or_delete`? I don't get you :)

solnic:
ugh, i'm tired...I meant `create or update`

flash-gordon:
That's what I thought, just decided to make it clear 😁  I haven't worked with PG's upserts yet, only with Oracle's `merge` operator which is waaaay more complex (and not atomic in contrast).

Anyway, according to the [docs](https://www.postgresql.org/docs/9.5/static/sql-insert.html) there are following options for `insert_on_confict`:
1. Action. It can be `do nothing` or `do update`. I think it is not needed to be set explicitly, the fact that we have (not) an update statement is sufficient for choosing a proper option.
2. Constraint name. I think this _should_ be configurable, though this is not a mandatory parameter.
3. Conflict target. This is an optional set of indexed columns that can be set _instead of_ constraint name. I'm not sure why and when this option should be used, but I guess this can be helpful when constraint name is unknown, ie auto-generated or sutin. We _could_ support for this option, the corresponding Sequel's parameter named `:target`.
4. Update. This is a little bit tricky. You can provide a static value for it, this means you'll be able to update an existing row having access to both existing row and so called _excluded_ row in SQL. It looks like you provide a transformation for the existing row expressed via a small piece of SQL. You can also provide this update statement for every `upsert` call be I doubt you'll ever need to do so. I think static configuration is good enough, and we _should_ allow to configure it.
5. Where clause for update part. We _could_ support for this but this would be used very occasionally fwiw. 

Sequel's docs http://sequel.jeremyevans.net/rdoc-adapters/classes/Sequel/Postgres/DatasetMethods.html#method-i-insert_conflict

lib/rom/sql/commands/postgres.rb
```diff
@@ -0,0 +1,20 @@
+module ROM
+  module SQL
+    module Commands
+      module Postgres
```
flash-gordon: I believe this should be named PG to be consistent with ROM::SQL::Types::PG

flash-gordon: Though the next file named `lib/rom/sql/commands_ext/postgres.rb` :)

flash-gordon:
I added support for all Sequel's (ie PG's) options you can provide for `insert on conflict` statement. Though I'm not confident with the API, I added specs for these options and everything works smooth so far.
/cc @gotar @solnic

flash-gordon:
I fixed specs for `Create` command (oops). It's worth mentioning that PG does not return anything on `ON CONFLICT DO NOTHING` and we can't find out what record exactly rejected the insert. I have no idea how to handle this case, probably the end user will need to update some dummy column to force PG to return the conflicted row.

flash-gordon:
Now when the specs use `insert on conflict` they are failing on Travis because it uses PG 9.1 by default. Using 9.5 on Travis is not so easy https://github.com/travis-ci/travis-ci/issues/4264, we either should not run that specs on PG < 9.5 or run them against 9.5 on Travis. The later option could slow down the build significantly so this requires some research beforehand, I'm going to do it today or tomorrow.

solnic:
@flash-gordon circle ci supports PG 9.5 OOTB maybe we should move there

flash-gordon:
@solnic not a bad idea, I use circle ci and I can say it is nice. I'll try to setup the environment for my fork first

solnic:
@flash-gordon yeah I've been using CircleCI a lot, it's super cool. Setup should be straight-forward. IIRC the provide parallel builds for OSS projects too, which could significantly speed up our builds if we split specs per adapter :) I enhanced our setup with ability to run specs for a specific adapter on-demand, it's not controlled by external variables yet though, but this could be easily done so that we could do stuff like `ADAPTER=sqlite rake spec`

gotar:
sequel support (except pgsql) mysql and sqlite if I good remember. So it should be very easy to extend this functionality to those databases too.

spec/integration/commands/upsert_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+        expect(command.call(excluded)).to be nil
+      end
+    end
+
+    context 'on conflict do update' do
+      context 'with conflict target' do
+        let(:command_config) do
+          -> do
+            conflict_target :title
+            update_statement user_id: 2
```
solnic: Does it mean it can't figure out the update_statement at runtime?

flash-gordon: @solnic the only reason why this is not possible is that command interface is lack of a method that builds a new command with new options
```ruby
          command_with_update = command.with_options(
            update_statement: {
              user_id: :excluded__user_id
            }
          )
```
But anyway, you are veeery unlikely need to set update_statement at runtime, that's why this is a class level option by default.

solnic:
@flash-gordon anything else left to do here?

solnic:
It looks good and it's a great starting point, I suspect we're gonna need to make some improvements to make it work smoothly with repos, but that can be done separately.

Thanks guys :)

PR #68, opened 2016-06-21
Add PG bytea type for binary data

flash-gordon:


solnic:
the CI failed and it looks similar to what I've heard lately that latest sequel version doesn't set logger the way it used to so we end up with `nil` :/

flash-gordon:
@solnic don't worry, I'll handle it 😎

flash-gordon:
@solnic is it ready to be merged?

solnic:
👍

Issue #67, opened 2016-06-08
Clarify schema syntax for defining associations 


daniels:
If a relation is registered under a different name than the underlying dataset, it is not obvious what name goes where when defining associations.

```ruby
config.relation(:users) do
  dataset :user

  schema(:user) do                                         # <= dataset name
    attribute :user_id, ROM::SQL::Types::Serial
    attribute :name, ROM::SQL::Types::String

    associate do
      many :tasks                                          # <= relation name
    end
  end
end

config.relation(:tasks) do
  dataset :task

  schema(:task) do                                         # <= dataset name
    attribute :task_id, ROM::SQL::Types::Serial
    attribute :user_id, ROM::SQL::Types::ForeignKey(:user) # <= dataset name
    attribute :name, ROM::SQL::Types::String

    associate do
      one :users                                           # <= relation name
    end
  end

  def for_users(users)
    where(user_id: users.map { |user| user[:user_id] })
  end
end
```

When mixing these up and using them (through `ROM::Repository#aggregate`) you get `NoMethodError` on `nil:NilClass`, where the missing method is `foreign_key` if you used relation name in `ROM::SQL::Types::ForeignKey()` and `primary_key` if you used dataset name in the `associate` block.

I'm not sure what the best solution would be here – changing the behavior or just emphasis the current behavior in the documentation. Or maybe it would help if the error was raised when the relation/dataset couldn't be found rather than when calling a method on the missing object.

flash-gordon:
OK, now we can say that we _always_ use database names. This means if you want to associate a relation with a name different than the underlying table name you go with
```ruby
associate do
  many :my_table_name, relation: :my_relation_name
end
```
Also you can provide an optional association name that can be used later in rom-repo
```ruby
schema(:posts) do
  attribute :author_id, ROM::SQL::Types::ForeignKey(:users) 

  associate do
    belongs :users, relation: :people, as: :owner
  end
end
###
people_repo.posts.combine(:owner).one.owner
```

PR #70, opened 2016-07-04
Use qualified combination keys

flash-gordon:
This leverages new relation name object (https://github.com/rom-rb/rom/pull/347) and also adds qualified identifiers for associations. Now it is possible to use associations when your PK/FK do not follow common naming conventions. I.e. something like `tenants.tenant_id (PK) <= users.tenant_id (FK)` should work fine.

This PR also contains a breaking change for ad hoc combination names in associations (this won't work anymore: https://github.com/rom-rb/rom-repository/blob/master/spec/shared/relations.rb#L58). Now when you write `belongs :sutin` `sutin` means a dataset (table) name, not a relation name. You can add a relation name with `belongs :weird_name, relation: :nice_name`. This should remove the confusion described in https://github.com/rom-rb/rom-sql/issues/67 We add a combination name later quite easily with a more verbose syntax eg `belongs :users, relation: :authors, as: :author`.

Current state: not ready, there are still some issues with combining. I also plan to add a bunch of specs for weird naming, this will guarantee support for legacy DB schemas.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,13 +4,13 @@ gemspec
 
 gem 'dry-types', github: 'dry-rb/dry-types', branch: 'master'
 gem 'rom', github: 'rom-rb/rom', branch: 'add-explicit-relation-name'
-gem 'rom-support', github: 'rom-rb/rom-support', branch: 'call-to_sym'
+gem 'rom-support', github: 'rom-rb/rom-support', branch: 'master'
 
 group :test do
   gem 'byebug', platforms: :mri
   gem 'anima', '~> 0.2.0'
   gem 'virtus'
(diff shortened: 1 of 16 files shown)


flash-gordon:
Travis fails because of AS :)

daniels:
I'm not sure, but https://github.com/rom-rb/rom-repository/blob/master/spec/shared/relations.rb#L58 seems to be for the use case where you want to give the *association* a different name from both the dataset and the relation. I.e. you might have a `user` (singular) table in a `users` (plural) relation, but from the `posts` relation you want to refer to it as `post.authors`. Could/Would this still be supported when the dataset/relation name issues are resolved?

PR #71, opened 2016-07-12
Refactor :associates plugin to use Association API

solnic:
This makes `associates` plugin use defined schema associations and delegate setting FKs. As a bonus commands now support many-to-many where join-tuples are automatically inserted for you.

Passing ad-hoc key options to `associates` macro is now deprecated.

### DIFF
--- a/lib/rom/sql/plugin/associates.rb
+++ b/lib/rom/sql/plugin/associates.rb
@@ -1,3 +1,5 @@
+require 'rom/support/deprecations'
+
 module ROM
   module SQL
     module Plugin
@@ -98,6 +100,14 @@ module ROM
             if associations.map(&:first).include?(name)
               raise ArgumentError,
                 "#{name} association is already defined for #{self.class}"
+            end
+
+            if options.any?
+              Deprecations.warn(
+                "Passing options to `associates` is deprecated. " \
+                "Define an association in schema for your :#{relation} relation " \
+                "instead and pass in its name to `associates`"
+              )
             end
 
             option :association, reader: true, default: {}
--- a/lib/rom/sql/plugin/associates.rb
+++ b/lib/rom/sql/plugin/associates.rb
@@ -14,17 +14,19 @@ module ROM
         end
 
         module InstanceMethods
+          attr_reader :assoc, :__registry__
+
           # @api private
           def initialize(*)
             super
-            self.class.associations.each do |(name, opts)|
-              association[:key] =
-                if opts.any?
-                  opts[:key]
-                else
-                  relation.schema.associations[name].join_key_map(relation.__registry__)
-                end
-            end
+            @__registry__ = relation.__registry__
+            assoc_name, assoc_opts = self.class.associations[0]
+            @assoc =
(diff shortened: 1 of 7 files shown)


solnic:
@flash-gordon could you take a look? Still WIP. I'm wondering if we should deprecate ad-hoc key options in `associates` plugin in favor of association definitions in schemas and simply store association object in a command that was configured to do the association and just delegate this work to the association object instead of making that custom `execute` so complex. This should scale better I think.

lib/rom/sql/plugin/associates.rb
```diff
@@ -39,14 +35,33 @@ def initialize(*)
...
+                input_tuples = with_input_tuples(tuples).map { |tuple|
+                  tuple.merge(fk => parent.fetch(pk))
+                }
 
-            super(input_tuples)
+                super(input_tuples)
+              when Hash
+                new_tuples = super(tuples)
+
+                join_relation, keys = assoc.to_a[0]
```
flash-gordon: It's better to unwrap values once because they don't change
```ruby
when Hash
   super(tuples).tap do |new_tuples|
     join_relation, ((spk, sfk), (tfk, tpk)) = assoc.to_a[0]

     join_tuples = new_tuples.map do |tuple|
       { sfk => tuple.fetch(spk), 
         tfk => parent.fetch(tpk) }
     end
     
     join_relation.multi_insert(join_tuples)
   end
 end
```

spec/integration/plugins/associates_spec.rb
```diff
@@ -58,33 +59,98 @@
...
+
+      before do
+        configuration.relation_classes[1].class_eval do
+          schema do
+            attribute :id, ROM::SQL::Types::Serial
+            attribute :user_id, ROM::SQL::Types::ForeignKey(:users)
+            attribute :title, ROM::SQL::Types::String
+
+            associate do
+              belongs :users, as: :user
```
flash-gordon: we could singularize association name by default for one-to-many, wdyt?

flash-gordon: not sure though

solnic: Crossed my mind. It's gonna be super common to use a singular name so...what would be the downside if we did that? Can't think of anything.

solnic:
@flash-gordon yeah I'll deprecate ad-hoc key config

spec/integration/plugins/associates_spec.rb
flash-gordon: the only downside I see is that it would the only implicit behavior so far :)

solnic: This seems harmless, I think. In general we avoid implicit behaviors (aka magic) but in this case it seems like a pragmatic choice as this association will be commonly used for aggregates where you definitely want a singular name so...

flash-gordon: agreed

solnic:
@flash-gordon 🎉 merge? :)

flash-gordon:
👍

PR #73, opened 2016-07-14
AssociationSet and Association::Name

solnic:


solnic:
@flash-gordon no more wip, thoughts?

spec/support/helpers.rb
```diff
@@ -0,0 +1,9 @@
+module Helpers
+  def qualified_name(*args)
```
flash-gordon: qualified_attribute?

flash-gordon:
@solnic 👍  looks fine to me `¯\_(ツ)_/¯`

spec/support/helpers.rb
solnic: ah thanks, renamed :)

PR #75, opened 2016-07-24
Infer foreign keys

flash-gordon:
This does not support multicolumn foreign key but I bet this is a really rare beast.


flash-gordon:
CircleCI's fail is not relevant, this branch does not have `circle.yml` yet.

PR #79, opened 2016-08-19
One To One and One To Many #associate

jodosha:
screen shot 2016-08-19 at 09 45 03

Here's the feature 😉 . This is useful for `hanami-model` integration. 💚


solnic:
Thank you 💜

PR #84, opened 2016-09-06
Added support for PostgreSQL types: UUID, Array, JSONB and Money

jodosha:
/cc @solnic @flash-gordon


lib/rom/sql/schema/column_inferrer.rb
```diff
@@ -0,0 +1,82 @@
+require 'rom/sql/types/pg'
```
flash-gordon: these pg extensions should be opt-in, inferrer class can be set in relation, so we can use different inferrers for different dbs by calling `schema_inferrer ROM::SQL::Schema::PGInferrer` in relation class

lib/rom/sql/types/pg.rb
```diff
@@ -5,21 +5,49 @@ module ROM
   module SQL
     module Types
       module PG
-        Sequel.extension(:pg_json)
+        # UUID
+
+        UUID = Types::Strict::String.constrained(format: Regexp.new('^([0-9a-f]{8})-([0-9a-f]{4})-([0-9a-f]{4})-([0-9a-f]{2})([0-9a-f]{2})-([0-9a-f]{12})$'))
```
flash-gordon: parens can be omitted, also it's common to use `\A` and `\z` in ruby instead of `^` and `$` because of some security issues, see [this SO post](http://stackoverflow.com/questions/577653/difference-between-a-z-and-in-ruby-regular-expressions) for details

flash-gordon:
I'm also afraid that insertion of array values wouldn't work, at least this is what I came across today trying to use an array type in a manually maintained schema. I'll have a deeper look into this tomorrow

lib/rom/sql/types/pg.rb
jodosha: You're right. I'm for removing that constraint for a few reasons:

  * It only supports UUID v4
  * It requires to be maintained
  * The original reason why I introduced it, was to control input and avoid malformed SQL statements (see _problem 1_)
  * We don't have a UUID type in Ruby. Indeed, `SecureRandom.uuid` returns an instance of `String`. We can keep `Types::PG::UUID` as a `Types::Strict::String` alias.

#### Problem 1

While using UUID as primary key, when the input isn't an UUID I get the following error:

```ruby
repository = SourceFileRepository.new
found = repository.find('foo')

# => 
#   Sequel::DatabaseError: PG::InvalidTextRepresentation: ERROR:  invalid input syntax for uuid: "foo"
      #   LINE 1: ...", "updated_at" FROM "source_files" WHERE ("id" = 'foo')...
```

I would expose a standardized behavior for missing records. When a table uses serial primary keys, if I do `repository.find('foo')` it returns `nil`. Tables with UUIDs as primary keys should behave the same.

jodosha:
@flash-gordon 
> I'm also afraid that insertion of array values wouldn't work

You're right, it wraps the array like this: `('ruby')`, but it should be `{'ruby'}`.

lib/rom/sql/schema/column_inferrer.rb
jodosha: How so? Can you please show me? Thanks 👍

flash-gordon: something like this https://gist.github.com/flash-gordon/901b94a9af623952f1babca50d7b8cd9 /cc @solnic can have some thought on this too

lib/rom/sql/types/pg.rb
flash-gordon: I wonder what SQL gets executed when you run `repository.find('foo')` with serial pks :) This should raise an error fwiw

jodosha:
@flash-gordon It looks like `Dry::Types::Definition#call` doesn't get called. I also added debuggers to all the definitions of `pg_array` in Sequel: none of them gets called.

lib/rom/sql/schema/column_inferrer.rb
jodosha: But then the problem becomes to explicit setup it.

flash-gordon: @jodosha well, technically, this is a manageable thing to do—to load a proper extension based on what type of connection we've got, though it'll require some rework around the inferrer infrastructure. /cc @solnic

solnic:
We can give it a shot. The only potential problem we may face is lack of
some info in the moment when we need it, since connection setup is not
available globally, we'd have to provide it to the inferer when relations
are being created.

gotar:
what with other db systems (other that pg?) for me it should be some kind of plugin, you can require manually.

flash-gordon:
@gotar I'm working on this atm, going to have a preview this weekend

flash-gordon:
@jodosha I force pushed the branch onto your repo, now I'm going to move some stuff here and there, then I'll let you know when it's ready or if I get any questions. The main thing, you probably already know about, is that I want to wrap all PG-specific code into an automatically activated extension

lib/rom/sql/extensions/postgres/types.rb
```diff
@@ -5,21 +5,64 @@ module ROM
...
-               .new(Sequel::Postgres::JSONHash)
-               .constructor(Sequel.method(:pg_json))
+        JSONHash = Dry::Types::Definition
+                   .new(Sequel::Postgres::JSONHash)
+                   .constructor(Sequel.method(:pg_json))
 
-        JSON = Array | Hash
+        JSONOp = Dry::Types::Definition
+                 .new(Sequel::Postgres::JSONOp)
+                 .constructor(Sequel.method(:pg_json))
```
solnic: Why is it `:pg_json` method?

flash-gordon: See [Sequel's docs](https://github.com/jeremyevans/sequel/blob/40bae5c0426c2a2bf24e92b0a6d880dffb4fbac0/lib/sequel/extensions/pg_json_ops.rb). TLDR Sequel allows to pass various values as a JSON(B) value but they are must be wrapped with `pg_json` call beforehand.

solnic: Ah gotcha

solnic:
This looks good. Pls merge <3

flash-gordon:
🎉

PR #85, opened 2016-09-06
Run specs against all DBs

flash-gordon:
I'm going to add ability to cherry-pick certain DB specs via an env variable


flash-gordon:
Now you can override adapters list for a spec/example group. Specs which was skipped by around hook are marked as pending, I think that's OK, because usually this requires updating a spec for compatibility with a certain DB

flash-gordon:
@solnic it's ready! 🎉 
All specs are now tagged with a specific database adapter and can be filtered with `bundle exec rspec -t sqlite`
/cc @jodosha

PR #86, opened 2016-09-10
Safe error wrapper

jodosha:
This PR aims to fix two issues:
1. [Sometimes](https://travis-ci.org/hanami/model/jobs/158778577) Sequel raises a `Sequel::ConstraintViolation`, when a database constraint is violated. I can't reproduce this locally (Mac OS), but it happens on CI and in general on Linux.
2. `ROM::Sql::Commands::ErrorWrapper` uses internally `ERROR_MAP` with `#[]`. This is dangerous because in case of non-mapped errors, instead of raising a generic error it raises `nil`, which causes a `TypeError`. I changed it with `#fetch` which falls back to the generic error (`ROM::Sql::Error`).

### DIFF
--- a/lib/rom/sql/commands/error_wrapper.rb
+++ b/lib/rom/sql/commands/error_wrapper.rb
@@ -11,7 +11,7 @@ module ROM
         def call(*args)
           super
         rescue *ERROR_MAP.keys => e
-          raise ERROR_MAP[e.class], e
+          raise ERROR_MAP.fetch(e.class, Error), e
         end
 
         alias_method :[], :call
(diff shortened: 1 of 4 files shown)


jodosha:
@solnic @flash-gordon I've added a test to prove that the code works. Can you please review? Thanks! 😉

flash-gordon:
LGTM, @solnic?

solnic:
Yes this is a good change. Could you clean up commits though? There are some commits from master that are not needed here, this should be just one commit.

flash-gordon:
I force-pushed the branch, but now it fails because of https://github.com/codeclimate/ruby-test-reporter/commit/52760d13430865c0cfc3618fae09f3c815939c73 Need to update codeclimate integration master, I'll fix it today. Specs are passing locally

jodosha:
@solnic @flash-gordon Please make sure this will be part of the next release. Thanks! 😄

solnic:
That's the plan :)

Issue #88, opened 2016-09-11
Inferred types are over-strict


cflipse:
Adding `schema(:db, infer:true)` to an existing relation, and I get the following errors:

``` 
     Dry::Types::ConstraintError:
       Sun, 11 Sep 2016 08:50:45 -0400 violates constraints (type?(Time) failed)
```

the column itself is a standard postgres timestamp:

```
 created_at   | timestamp without time zone | 
```

solnic:
@flash-gordon wdyt about using non-strict types by default? and maybe adding a `strict: true` option to the inferrer?

Regardless of this, I wonder what the actual type is, since `type?(Time)` failed then...what the hell is it? :D

flash-gordon:
@solnic it's not a bad idea to make it configurable, but what kind of types are you suggesting instead? Coercible? This is doable, no doubt, but I'm not sure if this will help in this particular case, depends on what @cflipse is trying to insert/update

cflipse:
It's coming in from a rom-model form with `timestamps` set on input ...  so ... DateTime.

flash-gordon:
Aha! So, dry-t doesn't have a coercible time type, this means it can't transform DateTime to Time ootb. Should we do something about it, @solnic?

jodosha:
I agree on this. Schema hash should be more relaxed. 👍 

For instance:

```ruby
UserRepository.new.create(name: 'L', active: nil)
  # => raises "nil violates constraints (type?(FalseClass) failed)"
```

(I'm using `relation.schema_hash` to do data transformations)

solnic:
Yeah I agree, this shouldn't be strict by default when inferring

solnic:
@flash-gordon just plain definitions ie `Types::String` etc.

jodosha:
In my branch I tried with `Maybe::Optional`. First of all `Bool`, `Date` and `Time` aren't defined under that module. Is that intentional?

I defined them, but `Maybe` types replace `nil` with `None`, which causes an error in Sequel cause it can't represent `None` as a SQL literal.

solnic:
@jodosha *don't* use maybes for this, it is not worth the effort and adds additional overhead :)

jodosha:
Fine :)

Should Coercible be enough?

flash-gordon:
@solnic plain definitions will ignore actual values IIRC. But what about special PG types, should be values coerced to custom Sequel's types or not?

solnic:
[...]

PR #94, opened 2016-09-19
Use optional types for inferred schema

jodosha:
Closes https://github.com/rom-rb/rom-sql/issues/88

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## v0.8.1 to-be-released
+
+### Added
+
+* `Associations::{OneToMany,OneToOne}#associate` for merging FKs into child tuple (jodosha)
+
+[Compare v0.8.0...HEAD](https://github.com/rom-rb/rom-sql/compare/v0.8.0...HEAD)
+
 ## v0.8.0 2016-07-27
 
 ### Added
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,10 @@ source 'https://rubygems.org'
 
 gemspec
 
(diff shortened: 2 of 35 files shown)


solnic:
We do use `optional` types [here](https://github.com/rom-rb/rom-sql/blob/master/lib/rom/sql/schema/inferrer.rb#L54). This doesn't work for you?

jodosha:
@solnic @flash-gordon There are several cases that a `Strict` type won't fit:

  * `Date`, `DateTime`, and `Time` are often represented as strings in the database. Strict types expect as input the exact type they represent. So `ROM::Sql::Types::Date[Date.today.to_s]` raises an error. We should be more flexible.

  * Similarly `Decimal` blows up if we pass a float: `ROM::Sql::Types::Decimal[3.14]`.

  * The actual behavior is: when we use `NOT NULL` constraint, we get a `Strict` type. If we have `users.active boolean NOT NULL default false`, when we omit the input (`active`): `repository.create(name: 'Luca')` we'll get an error from ROM/dry-types (`Message: <"nil (NilClass) has invalid type for :active">`). Shall we let the database to handle this case?

flash-gordon:
I guess we fixed all the issues we've got so far, so this can be closed. However, there is a bunch of type specs in the PR, @solnic what do you think about them?

flash-gordon:
I guess we fixed all the issues we've got so far, so this can be closed. However, there is a bunch of type specs in the PR, @solnic what do you think about them?

jodosha:
Closing in favor of the already merged https://github.com/rom-rb/rom-sql/pull/101 👍

PR #95, opened 2016-09-22
add SQL having filter

cflipse:


cflipse:
```
SELECT "users"."id", "users"."name", count("tasks") AS "goal_count" FROM "users" LEFT OUTER JOIN "tasks" ON ("tasks"."user_id" = "users"."id") GROUP BY "users"."id", "users"."name" HAVING ("goal_count" = 2)
 ORDER BY "users"."id
```

That's the SQL generated in the failing test; claims not to know what "goal_count" is.

This is one of those cases where things _look_ right, but I must be missing _something_ ?

solnic:
SQL eh? :) I'll be looking into that tomorrow, I'd like to wrap up new releases for rom core/sql

solnic:
Here's a valid SQL:

``` sql
SELECT
  "users"."id", "users"."name", COUNT("tasks") AS "goal_count"
FROM "users"
LEFT OUTER JOIN "tasks" ON ("tasks"."user_id" = "users"."id")
GROUP BY "users"."id", "users"."name"
HAVING COUNT("tasks") > 1
ORDER BY "users"."id"
```

So the spec should be tweaked. In general specs for individual relation methods should be in `spec/unit/relation/#{method}_spec` so ie `spec/unit/relation/having_spec.rb` (and we gotta extract exisiting specs into separate files in master)

cflipse:
Got it.  I _think_ I just looked for where tests on `group` were, and added there.

Will update specs.

cflipse:
oh, I see ... `spec/unit/relation/{method}_spec` is aspirational, which is why I didn't find anything there ...

Hrm; short of string-comparing the SQL, or spies / mocks on Sequel, how should we structure those specs as _unit_ specs ?

solnic:
There's a continues effort to improve the test suite in terms of
file/example organization. The biggest challenge was to make it flexible
for multi-gateway testing (done already) and now we gotta focus on
splitting things up and improving coverage. The spec you used is considered
to be "old" from my pov and should be refactored :)

cflipse:
> HAVING COUNT("tasks") > 1

and, hrm ...  I'll cop to not-being-an-expert, but most of the time, I've been able to refer to column aliases in an SQL having clause ... is this one of those things that's not actually cross-db ?

solnic:
> and, hrm ... I'll cop to not-being-an-expert, but most of the time, I've been able to refer to column aliases in an SQL having clause ... is this one of those things that's not actually cross-db ?

it doesn't work like that *at least* on postgresql

solnic:
> Hrm; short of string-comparing the SQL, or spies / mocks on Sequel, how should we structure those specs as unit specs ?

That's a good question. Comparing sql strings is brittle, on the other hand executing queries and comparing results requires much more complicated setup, *but* it's a much more reliable way of testing that things work. If we want to be pedantic about test types, we can put those under `spec/integration`.

solnic:
This is now merged (I rebased so commits were rewritten so this wasn't automatically closed). It's gonna be in 0.9.0

PR #100, opened 2016-10-13
Make OneToOne inherit from OneToMany.

beauby:
Previously, the `OneToMany` association was inheriting from the `OneToOne`. This is mainly a cosmetic change.


solnic:
Thanks man :)

PR #101, opened 2016-10-13
Make inferrer to use non-strict types

flash-gordon:
See https://github.com/rom-rb/rom-sql/issues/88

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,8 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem 'dry-core', github: 'dry-rb/dry-core'
-gem 'rom', github: 'rom-rb/rom'
+gem 'dry-core', github: 'dry-rb/dry-core', branch: 'master'
+gem 'rom', github: 'rom-rb/rom', branch: 'master'
 
 group :test do
   gem 'byebug', platforms: :mri
--- a/lib/rom/sql/extensions/postgres/types.rb
+++ b/lib/rom/sql/extensions/postgres/types.rb
@@ -7,7 +7,7 @@ module ROM
       module PG
         # UUID
 
-        UUID = Types::Strict::String
+        UUID = Types::String
 
         # Array
 
@@ -62,7 +62,7 @@ module ROM
 
         # MONEY
 
-        Money = Types::Strict::Decimal
+        Money = Types::Decimal
       end
     end
   end
--- a/lib/rom/sql/schema/inferrer.rb
+++ b/lib/rom/sql/schema/inferrer.rb
@@ -8,13 +8,13 @@ module ROM
         defines :ruby_type_mapping, :numeric_pk_type, :db_type, :db_registry
 
         ruby_type_mapping(
-          integer: Types::Strict::Int,
-          string: Types::Strict::String,
-          date: Types::Strict::Date,
-          datetime: Types::Strict::Time,
-          boolean: Types::Strict::Bool,
-          decimal: Types::Strict::Decimal,
-          float: Types::Strict::Float,
+          integer: Types::Int,
+          string: Types::String,
+          date: Types::Date,
+          datetime: Types::Time,
+          boolean: Types::Bool,
+          decimal: Types::Decimal,
+          float: Types::Float,
           blob: Types::Blob
         ).freeze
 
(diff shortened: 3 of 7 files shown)


solnic:
Could you add a couple of specs that verify that it works, esp when a value can be implicitly coerced by sequel, and when there's `nil` (to make sure `optional` works as expected)?

flash-gordon:
yeah, I was thinking about it. Just not sure what exactly we need to test, there are various cases for missing, `nil`, and uncoerced values mixed with different DB datatypes, implicit coercion policies and `null/not null/default` options.  I'll spend some time investigating Sequel's behavior. Also do you think we need a strict inferrer or not (e.g. as an option)?

solnic:
@flash-gordon at least the basic scenarios like `user_repo.create(name: 'foo', age: nil)` where `age` is an optional. I actually have second thoughts about implicit coercions, since this way we'd start relying on sequel coercion (if it exists, actually) and that's not what I would like to have tbh. In general when a field has a type specified then only this type should be used, so ie if we have a date_time field in the db, then only DateTime objects should be allowed. We need to take a closer look at what Sequel is doing in such cases, and figure out whether we want to have the same behavior or not.

solnic:
One more thing - we're making this change as folks had specific problems with strict types, which means there were specific cases where it didn't work, so, let's start with spec examples covering these specific cases so that we know this change is fixing that for people. /cc @jodosha @cflipse

flash-gordon:
@solnic OK, so it seems Sequel doesn't do anything special with values unless they are has special Sequel types, which is good fwiw. All coercions are made by DB-specific adapters, not by Sequel itself. Have a look at [the specs I wrote](https://github.com/rom-rb/rom-sql/pull/101/files#diff-8ce4d60fa4157c7143b75b2a0cb11762R97). I got a really weird behavior for SQLite under JRuby, it  inserts strings instead of timestamps I've no idea why this happens

flash-gordon:
SQLite under JRuby writes data to the database as is, no coercions involved.

flash-gordon:
Also, I want to stress that if a column has a default value and a not null constraint `nil` won't be ignored so that a resulting `INSERT` statement will try insert `NULL` i.e. violate the constraint. _I think_ it's alright/expected, wdyt? /cc @solnic @jodosha @cflipse

cflipse:
Hey, sorry ... been a busy week.

So, if you're saying that if I try and insert a row with, say, `user_id: nil` and there's a not null constraint ... I'm totally okay with that erroring if I try to _set_ the column to null.

If it's an update, and the column isn't named in update parameters, than I wouldn't expect an error, but I'm pretty sure that's not what we're talking about here ... ;)

cflipse:
aaand, I tried to check the branch locally against the project I had that was inspiring the original bug report, but ...

```
flip@meteu:~/src/get_ready_server (check-rom)$ be rspec
bundler: failed to load command: rspec (/home/flip/.gem/ruby/2.3.0/bin/rspec)
Bundler::GemRequireError: There was an error while trying to load the gem 'rom-sql'.
Gem Load Error is: undefined method `define' for nil:NilClass
Backtrace for gem load error is:
/home/flip/src/rom-rb/rom-support/lib/rom/support/options.rb:165:in `option'
/home/flip/src/rom-rb/rom-sql/lib/rom/sql/association/many_to_many.rb:9:in `<class:ManyToMany>'
/home/flip/src/rom-rb/rom-sql/lib/rom/sql/association/many_to_many.rb:4:in `<class:Association>'
/home/flip/src/rom-rb/rom-sql/lib/rom/sql/association/many_to_many.rb:3:in `<module:SQL>'
```

that using a GH based checkout of `rom`, `rom-support` and `rom-sql` .. I'm too tired to dig further on that right now.  on the one hand, I know that using the new updates is helpful and occasionally necessary.  On the other hand, it occasionally makes checking a bugfix branch like this a real pain.  :/

solnic:
> Also, I want to stress that if a column has a default value and a not null constraint nil won't be ignored so that a resulting INSERT statement will try insert NULL i.e. violate the constraint. I think it's alright/expected, wdyt?

`nil` is a value, so if you pass in a nil, you pass a value, if a column has not-null constraint, it is expected that it will raise a constraint error. If somebody wants to rely on db-side default values, then nils should not be passed to commands.

PR #102, opened 2016-10-27
Add support for enum DB type

vladra:
I'm using native db `enum` type in my app and was having this error:

`ROM::SQL::UnknownDBTypeError: Cannot find corresponding type for enum`

This solved my problem. Please suggest if there should be any changes.

### DIFF
--- a/lib/rom/sql/schema/inferrer.rb
+++ b/lib/rom/sql/schema/inferrer.rb
@@ -15,7 +15,8 @@ module ROM
           boolean: Types::Bool,
           decimal: Types::Decimal,
           float: Types::Float,
-          blob: Types::Blob
+          blob: Types::Blob,
+          enum: Types::String
         ).freeze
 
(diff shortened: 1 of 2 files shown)


solnic:
Thanks :)

@flash-gordon wdyt about this?

flash-gordon:
Hey @vladra 
DB-specific types should go to corresponding inferrers, see [this one](https://github.com/rom-rb/rom-sql/blob/master/lib/rom/sql/extensions/postgres/inferrer.rb) as an example.
PG's enums require more work. First of all, `String :enum` creates a column with type `text` and name `enum`, not a column with `enum` type. See [sequel's sources](https://github.com/jeremyevans/sequel/blob/master/lib/sequel/extensions/pg_enum.rb) for how this should be fixed.  Also, I think we should add possible enum values to type's meta, something like this `type.meta(values: enum_values)`.

PR #105, opened 2016-11-07
Load db-extensions in gateway

flash-gordon:
Fixes one failure in https://github.com/hanami/model/tree/new-engine

### DIFF
--- a/lib/rom/sql/gateway.rb
+++ b/lib/rom/sql/gateway.rb
@@ -23,6 +23,10 @@ module ROM
         attr_accessor :instance
       end
 
+      CONNECTION_EXTENSIONS = {
+        postgres: %i(pg_array pg_json)
+      }.freeze
+
       # Return optionally configured logger
       #
       # @return [Object] logger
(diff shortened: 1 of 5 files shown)


jodosha:
@flash-gordon Thank you. 💚 It works locally! yay! 🎉

lib/rom/sql/gateway.rb
```diff
@@ -177,11 +181,18 @@ def connect(uri, *args)
...
-      def add_extensions(exts)
-        connection.extension(*exts)
+      def load_extensions(exts)
+        db_type = connection.database_type.to_sym
+
+        if ROM::SQL.available_extension?(db_type)
+          ROM::SQL.load_extensions(db_type)
+        end
+
+        extensions = (CONNECTION_EXTENSIONS.fetch(db_type) { [] } + exts).uniq
```
AMHOL: Any reason for using the block default with `#fetch` here over `CONNECTION_EXTENSIONS.fetch(db_type, [])`, just out of interest?

flash-gordon: That's a good question. I did it automatically tbh, but the difference is that `, [])` will always allocate create a new array and `{ [] }` won't. Should be `fetch(db_type, EMPTY_ARRAY)` instead of both variants :)

AMHOL: :+1:

PR #107, opened 2016-11-19
Timestamp plugin

cflipse:
Adding a plugin for configuring commands to auto-timestamp.

This could _probably_ go into ROM core, as there's nothing database specific about it, but I'd like to iron out any kinks in a place where it's ... nominally expected.

I'm writing out the time explicitly in UTC, but I'm not great with timezone issues, so if anyone wants to point out the spots where I'm being blind here, that'd be great.

### DIFF
--- a/spec/unit/plugin/timestamp_spec.rb
+++ b/spec/unit/plugin/timestamp_spec.rb
@@ -8,6 +8,7 @@ RSpec.describe 'Plugin / Timestamp' do
     before do
       conf.commands(:notes) do
         define :create do
+          result :one
           use :timestamps
           timestamp :updated_at, :created_at
         end
@@ -25,7 +26,9 @@ RSpec.describe 'Plugin / Timestamp' do
     end
 
     it "applies timestamps by default" do
-      pending
+      result = container.command(:notes).create.call(text: "This is a test")
+
+      expect(result).to include(:created_at).and(include(:updated_at))
     end
 
 
--- a/spec/unit/plugin/timestamp_spec.rb
+++ b/spec/unit/plugin/timestamp_spec.rb
@@ -13,6 +13,13 @@ RSpec.describe 'Plugin / Timestamp' do
           timestamp :updated_at, :created_at
         end
 
+        define :create_many, type: :create do
+          result :many
+          use :timestamps
+          timestamp :updated_at, :created_at
+        end
+
+
         define :update do
           use :timestamps
           timestamp :updated_at
@@ -29,6 +36,16 @@ RSpec.describe 'Plugin / Timestamp' do
       result = container.command(:notes).create.call(text: "This is a test")
 
       expect(result).to include(:created_at).and(include(:updated_at))
+    end
+
+    it "sets timestamps on multi-tuple inputs" do
+      input = [{text: "note one"}, {text: "note two"}]
+
+      results = container.command(:notes).create_many.call(input)
+
+      results.each do |result|
+        expect(result).to include(:created_at).and(include(:updated_at))
+      end
     end
 
 
--- a/spec/unit/plugin/timestamp_spec.rb
+++ b/spec/unit/plugin/timestamp_spec.rb
@@ -11,6 +11,7 @@ RSpec.describe 'Plugin / Timestamp' do
           result :one
           use :timestamps
           timestamp :updated_at, :created_at
+          datestamp :written
         end
 
         define :create_many, type: :create do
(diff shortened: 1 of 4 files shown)


cflipse:
So, it appears that this is failing CI on JRuby when run against SQLite, becase SQLite does not have a time datatype ... I presume that this _should_ be handled by a lower level adapter, and may be a bug in, say, sequel -- but I'm not sure it's really something that can be dealt with at that level?

I _could_ make the tests pass by just accounting for the case where the database fails to adapt the time _back_ into a Time object, but I feel like that might just be dodging a different problem...

cflipse:
Also:  I'm considering also adding a `datestamp` function, which sets the specified attributes to the current date:  `datestamp :completed_on`  ... 

alternatively, could do the much less intuitive approach of not-handling the dates automatically:   `datestamp(:completed_on){ Date.today }`  ... which, however, ends up feeling much more like a "default attribute fill" plugin than a timestamp plugin (and could conceivably end up with not-matching times on multiple stamps ...)  ... think I've talked myself out of the block ;)

solnic:
> So, it appears that this is failing CI on JRuby when run against SQLite, becase SQLite does not have a time datatype ... I presume that this should be handled by a lower level adapter, and may be a bug in, say, sequel -- but I'm not sure it's really something that can be dealt with at that level?

Sequel doesn't convert this for you because sqlite doesn't support time/date column types:

```
irb(main):010:0> db = Sequel.connect('jdbc:sqlite::memory')
=> #<Sequel::JDBC::Database: "jdbc:sqlite::memory">
irb(main):011:0> db.create_table(:notes) { primary_key :id; column :created_at, DateTime }
=> nil
irb(main):012:0> db[:notes].insert(created_at: DateTime.now)
=> 1
irb(main):013:0> db[:notes].first
=> {:id=>1, :created_at=>"2016-12-06 15:02:47.056000"}
irb(main):014:0> Note = Class.new(Sequel::Model(:notes))
=> Note
irb(main):015:0> Note.first
=> #<Note @values={:id=>1, :created_at=>"2016-12-06 15:02:47.056000"}>
```

cflipse:
So, why does it pass on rubies not named `jquery` ?

solnic:
They use a different driver (jruby uses jdbc-sqlite3)

cflipse:
hah!  So it _is_ a driver thing, just much lower level. ;)

I'll tweak the spec to check against a string version.

solnic:
Yeah it is, though I gotta say this discrepancy in behavior is pretty bad, but I'm not going to worry about it (at least not now) :)

cflipse:
Fixed spec, added a `datestamp`

spec/integration/plugins/timestamp_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+  include_context 'relations'
+
+  with_adapters do
+    context "with Create command" do
+
+    end
+
+  end
+
+end
```
solnic: I suppose this can be deleted?

spec/unit/plugin/timestamp_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      expect(result).to include(:created_at).and(include(:updated_at)).and(include(:written))
+    end
+
+    it "sets timestamps on multi-tuple inputs" do
+      input = [{text: "note one"}, {text: "note two"}]
+
+      results = container.command(:notes).create_many.call(input)
+
+      results.each do |result|
+        expect(result).to include(:created_at).and(include(:updated_at))
```
solnic: Does this check if these values are filled?

cflipse: Mostly checks that the keys are present, which implies that ... _something_ has been done.  But, no, doesn't verify the times are filled.  dodging time comparisons mostly.

Will update.

solnic: Yeah these keys will always be there, so it's important to check if we get back date time instances :)

spec/integration/plugins/timestamp_spec.rb
cflipse: Er, yep.

I'm not actually sure about the split between Unit an Integration in these cases; I ended up following one of the _other_ plugin specs and building into unit, but ... I tend to feel like things that actually hit the database a Really Integration Tests.  

:shrug:

spec/unit/plugin/timestamp_spec.rb
cflipse: Did the parsing song & dance, and checked that the time is correct within a second.  that _should_ be good ... I'm hesitant to want to drag in something like `timecop` _just_ so one can do time comparisons.  :roll_eyes:

cflipse:
will rework this to use the `before` methods

solnic:
Closing in favor of #135

PR #111, opened 2016-12-04
Add Relation#read(sql_str)

solnic:
refs hanami/model#345
refs hanami/model#354

### DIFF
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -507,6 +507,20 @@ module ROM
         def unique?(criteria)
           where(criteria).count.zero?
         end
+
+        # Return a new relation from a raw SQL string
+        #
+        # @example
+        #   users.read('SELECT name FROM users')
+        #
+        # @param [String] sql The SQL string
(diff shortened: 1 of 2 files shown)


solnic:
@flash-gordon so I did this, because it should be useful in general + hanami folks will benefit. Not entirely sure if the method name is good though

flash-gordon:
JFYI, in Sequel it's called "fetch" (`#[]` == `#fetch`), in my opinion it's not a fortunate name either

timriley:
`#read` actually seems pretty good to me. It feels analogous with reading a file, both in terms of potential cost and intended result.

solnic:
@flash-gordon yeah both `#[]` and `#fetch` are very bad names. We actually already have `#fetch` that expects PK and returns a single tuple

solnic:
FWIW in rom-yesql the same method (but in Dataset) is also [called read](https://github.com/rom-rb/rom-yesql/blob/master/lib/rom/yesql/dataset.rb#L22-L24). So let's just stick to it.

PR #116, opened 2016-12-21
Support for inferring PG enum types

flash-gordon:


lib/rom/sql/extensions/postgres/inferrer.rb
```diff
@@ -7,17 +8,11 @@ class Schema
...
-          'numeric'          => true,
-          'real'             => true,
-          'double precision' => true,
-          'serial'           => true,
-          'bigserial'        => true,
-        ).freeze
+        db_numeric_types %w(
+          smallint integer bigint
+          decimal numeric real
+          double\ precision serial bigserial
```
AMHOL: Cool, learned something new :+1:

flash-gordon: yep, same here, I tried and it worked :tada: :)

solnic:
@flash-gordon well, that was fast mr Flash <3

PR #118, opened 2016-12-28
Expose schema attributes in restriction blocks

solnic:
This adds access to schema attribute API in `where` and `having` blocks

### DIFF
--- a/lib/rom/sql/expression.rb
+++ b/lib/rom/sql/expression.rb
@@ -1,0 +1,26 @@
+module ROM
+  module SQL
+    class Expression
+      attr_reader :expr, :type
+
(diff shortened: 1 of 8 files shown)


lib/rom/sql/expression.rb
```diff
@@ -0,0 +1,26 @@
...
+        @expr = expr
+      end
+
+      def sql_literal(ds)
+        expr.sql_literal(ds)
+      end
+
+      private
+
+      def method_missing(meth, *args, &block)
```
AMHOL: Should also define `respond_to_missing?`

solnic: Done :)

PR #119, opened 2016-12-29
Add support for setting custom FKs in associations

solnic:
### TODO

- [x] belongs_to
- [x] has_many
- [x] has_one
- [x] has_many-through

### DIFF
--- a/lib/rom/sql/association/many_to_one.rb
+++ b/lib/rom/sql/association/many_to_one.rb
@@ -15,6 +15,8 @@ module ROM
           left_schema = left.schema
           right_schema = right.schema.project_pk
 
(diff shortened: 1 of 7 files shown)


lib/rom/sql/association/many_to_one.rb
```diff
@@ -15,6 +15,8 @@ def call(relations)
           left_schema = left.schema
           right_schema = right.schema.project_pk
 
+          join_fk = foreign_key || right.foreign_key(target.relation)
```
flash-gordon: unused variable?

solnic: ugh right, this is what happens when you don't run all specs after making changes, thanks :) fixing...

PR #122, opened 2017-01-02
Update to use dry-initializer instead of ROM::Options

solnic:
### TODO

- [x] port associations
- [x] port pagination plugin
- [x] figure out a way to filter out non-sequel connection options
- [x] figure out a way to make `Associates` plugin work, as it needs to overwrite `initialize` method which is not supported by dry-initializer (yet)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'rom', git: 'https://github.com/rom-rb/rom.git', branch: 'master'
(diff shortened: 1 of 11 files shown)


solnic:
@nepalez @flash-gordon mostly done, but two things are missing. It would be good if we could override `initialize` in some cases and just call `super`

flash-gordon:
@solnic I added classmacros removal to the branch, placed your wip commit on top. Need this for rom-repo

solnic:
Great! There are still some issues with migrator but dry-initializer 0.11 works like a charm with overridden initialize :)

Cheers
Piotr

PR #124, opened 2017-01-05
Merge on to new Hash to simplify logic

parndt:
Rather than merging on top of options and then referencing its initial value for `as`,
we can simply define the default value inside a new `Hash`, and then merge the user
supplied options on top of it.


solnic:
Thanks man :)

PR #125, opened 2017-01-05
Raise a more helpful error when join table can't be found

georgemillo:
I was trying to create some joins and kept getting this cryptic error:

```
NoMethodError: undefined method `[]' for nil:NilClass
  from /Users/george/.rvm/gems/ruby-2.3.3/bundler/gems/rom-sql-c46d3c34c071/lib/rom/sql/schema.rb:58:in `join'
  from /Users/george/.rvm/gems/ruby-2.3.3/bundler/gems/rom-sql-c46d3c34c071/lib/rom/sql/relation/reading.rb:541:in `block in __join__'
  from /Users/george/.rvm/gems/ruby-2.3.3/bundler/gems/rom-sql-c46d3c34c071/lib/rom/sql/relation/reading.rb:541:in `each'
  from /Users/george/.rvm/gems/ruby-2.3.3/bundler/gems/rom-sql-c46d3c34c071/lib/rom/sql/relation/reading.rb:541:in `reduce'
  from /Users/george/.rvm/gems/ruby-2.3.3/bundler/gems/rom-sql-c46d3c34c071/lib/rom/sql/relation/reading.rb:541:in `__join__'
  from /Users/george/.rvm/gems/ruby-2.3.3/bundler/gems/rom-sql-c46d3c34c071/lib/rom/sql/relation/reading.rb:436:in `left_join'
```

turns out that I was doing it wrong: `inner_join(:task, user_id: :id)`, when
actually I should have passed `:tasks` as the first argument (i.e. the name of
the table.) An easy mistake to make since my user actually only has one task,
so I'm used to thinking about 'user.task' in the singular, not plural.

This PR raises a more helpful error when you get the name of the table wrong...
except it doesn't actually work, because `ROM::SQL::Error` has some extra magic
with `original_exception` that prevents `raise ROM::SQL::Error, 'message'` from
working as you'd expect.

Before I finalise this I figured I'd open the PR and a) make sure this is
actually a change you'd accept and b) get some feedback on whether I'm better
off raising ROM::SQL::Error (with some tweaks or hacks to make it actually
work) or using a different error class altogether.

### DIFF
--- a/lib/rom/sql/schema.rb
+++ b/lib/rom/sql/schema.rb
@@ -55,7 +55,7 @@ module ROM
 
       # @api public
       def join(name)
-        schema = relations.detect { |_, r|
+        schema = relations.detect(method(:join_table_not_found)) { |_, r|
           r.schema.name.dataset == name.to_sym
         }[1].schema
 
@@ -91,6 +91,12 @@ module ROM
           @primary_key_name = primary_key[0].meta[:name]
           @primary_key_names = primary_key.map { |type| type.meta[:name] }
         end
+      end
+
+      private
+
+      def join_table_not_found
+        raise ROM::SQL::Error, "can't join with '#{name}'; table not found"
       end
     end
   end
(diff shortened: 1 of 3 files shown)


solnic:
Thanks, I actually thought about doing something else - it feels like it'd be better to actually accept relations as arguments in `*_join` methods, so `inner_join(tasks)` instead of `inner_join(:tasks)`. This way the lookup in schema is just `relations[relation.name]`. This has nice benefit that aliases will automatically work, you people won't have to think about whether they need to use dataset name or relation name (in case it's the same). WDYT?

solnic:
I ended up removing this whole thing, so I need to close this. Thanks anyway :)

PR #126, opened 2017-01-06
Refactor associates plugin to use hooks

solnic:
refs rom-rb/rom#374

### DIFF
--- a/lib/rom/sql/plugin/associates.rb
+++ b/lib/rom/sql/plugin/associates.rb
@@ -7,51 +7,36 @@ module ROM
       module Associates
         # @api private
         def self.included(klass)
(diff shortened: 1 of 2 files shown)


lib/rom/sql/plugin/associates.rb
```diff
@@ -61,22 +46,48 @@ def execute(tuples, parent)
...
         module ClassMethods
-          # @api private
-          def inherited(klass)
-            klass.defines :associations
-            klass.associations []
-            super
+          # @see ROM::Command::ClassInterface.build
+          #
+          # @api public
+          def build(relation, options = EMPTY_HASH)
```
flash-gordon: not a trivial piece of code heh

PR #128, opened 2017-01-08
Add ROM::SQL::Gateway#transaction

flash-gordon:
Now we don't need to expose the connection object for that.


Gemfile
```diff
@@ -2,7 +2,7 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem 'rom', git: 'https://github.com/rom-rb/rom.git', branch: 'master'
+gem 'rom', git: 'https://github.com/rom-rb/rom.git', branch: 'feature/transactions'
```
solnic: I merged PR in rom already so this can be `master` again :)

Issue #129, opened 2017-01-11
Add support for missing types in inferrers


solnic:
There are some PG-specific types that we should add to [pg inferrer](https://github.com/rom-rb/rom-sql/blob/master/lib/rom/sql/extensions/postgres/inferrer.rb). This is especially important for hanami-model where it's not possible to define attributes manually (at least not yet), so **this is a blocker for hanami-model 1.0.0**.

### TODO

Specs should go to `spec/integration/schema/inferrer/#{db_type}_spec`, we have a bunch of specs already for common stuff, if there's anything db-specific in existing specs, it should be moved to a new spec that's dedicated to a given db engine.

- [x] Add `Types::PG::IPAddress` (this can be a `Types::String` for a good start)
- [x] Add `Types::PG::GeoPoint` (I'm not exactly sure what this should be :))
- [x] Add `Inferrer[:mysql]`
- [x] Add `Inferrer[:sqlite]`
- [x] Extend PG inferrer with a mapping `"inet" => Types::PG::IPAddress`
- [x] Extend PG inferrer with a mapping `"point" => Types::PG::GeoPoint`
- [x] Extend inferrers with a mapping `"numeric(precision)" => Types::Float.meta(precision: precision)` (I think?)
- [x] Extend inferrers with a mapping `"decimal(precision)" => Types::Decimal.meta(precision: precision`
- [x] Extend inferrers with a mapping `"decimal(precision, scale)" => Types::Decimal.meta(precision: precision, scale: scale)`
- [ ] Figure out which numeric/decimal db types are common for sqlite/mysql/postgres, ie from what I've seen `numeric(precision)` is used by Sequel in all 3 dbs

/cc @flash-gordon @jodosha

jodosha:
@solnic Thanks for defining all the missing steps. 👍

PR #131, opened 2017-01-14
Add missing PG types with auto-inference

flash-gordon:
Closes https://github.com/rom-rb/rom-sql/issues/129

### DIFF
--- a/lib/rom/sql/extensions/postgres/commands.rb
+++ b/lib/rom/sql/extensions/postgres/commands.rb
@@ -12,16 +12,14 @@ module ROM
           def insert(tuples)
             tuples.map do |tuple|
               relation.dataset.returning(*relation.columns).insert(tuple)
-            end.flatten.map { |t| relation.output_schema[t] }
+            end.flatten(1)
           end
 
           # Executes multi_insert statement and returns inserted tuples
           #
           # @api private
           def multi_insert(tuples)
-            relation.dataset.returning(*relation.columns).multi_insert(tuples).map { |t|
-              relation.output_schema[t]
-            }
+            relation.dataset.returning(*relation.columns).multi_insert(tuples)
           end
 
           # Executes upsert statement (INSERT with ON CONFLICT clause)
@@ -29,9 +27,7 @@ module ROM
           #
           # @api private
           def upsert(tuple, opts = EMPTY_HASH)
-            relation.output_schema[
-              relation.dataset.returning(*relation.columns).insert_conflict(opts).insert(tuple)
-            ]
+            relation.dataset.returning(*relation.columns).insert_conflict(opts).insert(tuple)
           end
         end
 
@@ -40,9 +36,7 @@ module ROM
           #
           # @api private
           def update(tuple)
-            relation.output_schema[
-              relation.dataset.returning(*relation.columns).update(tuple)
-            ]
+            relation.dataset.returning(*relation.columns).update(tuple)
           end
         end
 
@@ -101,7 +95,7 @@ module ROM
               upsert(input[tuple], upsert_options)
             end
 
-            inserted_tuples.flatten.map { |t| relation.output_schema[t] }
+            inserted_tuples.flatten
           end
 
           # @api private
--- a/lib/rom/sql/extensions/postgres/commands.rb
+++ b/lib/rom/sql/extensions/postgres/commands.rb
@@ -12,7 +12,7 @@ module ROM
           def insert(tuples)
             tuples.map do |tuple|
               relation.dataset.returning(*relation.columns).insert(tuple)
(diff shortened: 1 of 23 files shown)


solnic:
Hey so I added support for read types so that we can coerce to whatever the hell we want when reading relation data (https://github.com/rom-rb/rom/pull/377)

Would that be enough to support these custom types?

lib/rom/sql/extensions/postgres/types.rb
```diff
@@ -56,6 +57,10 @@ def self.Array(db_type)
                 .new(Sequel::SQL::Blob)
                 .constructor(Sequel::SQL::Blob.method(:new))
 
+        IPAddress = Dry::Types::Definition
+                    .new(IPAddr)
+                    .constructor(IPAddr.method(:new))
```
solnic: I added a shortcut for that:

``` ruby
require 'rom/types'

ROM::Types.Constructor(IPAddr, &IPAddr.method(:new))
```

flash-gordon:
@solnic this is awesome, I'll give it a try and we'll see whether it's enough

Also, I added precision/scale inferring for decimal/numeric types. I've found that decimal == numeric, at least from a practical point of view, so they both are mapped to `Decimal`

flash-gordon:
SQL2003:
```
 <exact numeric type> ::=
    NUMERIC [ <left paren> <precision> [ <comma> <scale> ] <right paren> ]
  | DECIMAL [ <left paren> <precision> [ <comma> <scale> ] <right paren> ]
  | DEC [ <left paren> <precision> [ <comma> <scale> ] <right paren> ]
  | SMALLINT
  | INTEGER
  | INT
  | BIGINT

 ...

21) NUMERIC specifies the data type
    exact numeric, with the decimal
    precision and scale specified by the
    <precision> and <scale>.

22) DECIMAL specifies the data type
    exact numeric, with the decimal scale
    specified by the <scale> and the
    implementation-defined decimal
    precision equal to or greater than the
    value of the specified <precision>.
```
but no one really implements this difference^, every RDBMS I know goes with NUMERIC, and DECIMAL is just an alias

solnic:
> I've found that decimal == numeric, at least from a practical point of view, so they both are mapped to Decimal

This is absolutely fine

lib/rom/sql/extensions/postgres/commands.rb
```diff
@@ -12,22 +12,26 @@ module Create
           def insert(tuples)
             tuples.map do |tuple|
               relation.dataset.returning(*relation.columns).insert(tuple)
-            end.flatten
+            end.flatten.map { |t| relation.read_schema[t] }
```
flash-gordon: @solnic we have to do this because command result must piped through `read_schema`

flash-gordon:
@solnic I think this is it, have a look. This PR depends on bi-directional mapping in ROM (https://github.com/rom-rb/rom/pull/377)

lib/rom/sql/extensions/postgres/commands.rb
```diff
@@ -12,22 +12,26 @@ module Create
...
           end
 
           # Executes multi_insert statement and returns inserted tuples
           #
           # @api private
           def multi_insert(tuples)
-            relation.dataset.returning(*relation.columns).multi_insert(tuples)
+            relation.dataset.returning(*relation.columns).multi_insert(tuples).map { |t|
+              relation.output_schema[t]
+            }
```
solnic: wow you already updated this to use `output_schema` OOOK :D

anyhow, I *think* it would be better to use an after hook for that, something like this:

``` ruby
module ROM
  module SQL
    class Create < ROM::Commands::Create
      after :finalize

      def finalize(tuples, *)
        with_input_tuples(tuples).map { |t| relation.output_schema[t] }
      end
    end
  end
end
```

Should be more flexible like that. WDYT?

flash-gordon: heh, yeah, hooks, I didn't become handy with them yet >.< I'll give it a try, but why do we need to call `with_input_tuples` in `finalize`, those are _output_ tuples, aren't they?

solnic: @flash-gordon `with_input_tuples` is just a convenient method that yields tuples even if it's *a tuple* (not an array with tuples). We could just rename it (and maybe even move to core, IIRC it's only in rom-sql but could be in core)

flash-gordon: @solnic https://github.com/rom-rb/rom-sql/pull/131/commits/df1cc691c7c0de8ddf720930c9dd0f8fd4df2a9c works like a charm ❤️

solnic:
@flash-gordon outstanding work man

PR #153, opened 2017-02-12
Make specs Oracle-compatible

flash-gordon:
The changes doesn't affect the code, only the specs have been touched. I split database setup into smaller files so that ROM won't infer relations for tables not used in a test. All the specs now cherry-pick tables and seeds according to their needs. The reason behind this change is a very slow dictionary lookup in Oracle. The operation takes so long that even after my changes it requires ~10 minutes waiting for all tests to complete.

Also I didn't run specs for Oracle under JRuby yet, will do it a bit later.


flash-gordon:
@solnic good news, this sped up tests for previously existed DBs by 1⁄3 🎉 I think I'll be able to make them even faster, we'll see.

solnic:
@flash-gordon that was another TODO on my "mental list"...it's really great you took care of it :)

PR #154, opened 2017-02-13
Fix the usage of timestamp in command chains

### LINKED ISSUE #152: Timestamp plugin breaks with chained commands
When a command has been chained, methods called with input tuples receive all of those tuples, which it does not currently support.

this results in an argument error:

```ruby
ArgumentError - wrong number of arguments (given 2, expected 1):                                                                                                                                         
  rom-sql (1.0.1) lib/rom/sql/plugin/timestamps.rb:99:in `set_timestamps'                                                                                                                                
  rom (3.0.1) lib/rom/command.rb:452:in `block in apply_hooks'                                                                                                                                           
  rom (3.0.1) lib/rom/command.rb:447:in `each'                                                                                                                                                           
  rom (3.0.1) lib/rom/command.rb:447:in `reduce'                                                                                                                                                         
  rom (3.0.1) lib/rom/command.rb:447:in `apply_hooks'                                                                                                                                                    
  rom (3.0.1) lib/rom/command.rb:274:in `call'
```

I'll get some specs around this and try to fix later this evening.

cflipse:
When triggered as a part of a command chain, `before` methods are called
with multiple arguments.  The initial implementation of `timestamp` as a
before filter did not take this into account.

```ruby
  create_task = rom.commands(:tasks).create_with_user
  create_user = rom.commands(:users).create.with(name: "John")

  (create_user >> create_task).call(name: "do the dishes")

  # ArgumentError:
  #   wrong number of arguments (given 2, expected 1)
  # ./lib/rom/sql/plugin/timestamps.rb:99:in `set_timestamps'
```

This fixes the filter method to address this case

Fixes #152

### DIFF
--- a/lib/rom/sql/plugin/timestamps.rb
+++ b/lib/rom/sql/plugin/timestamps.rb
@@ -96,7 +96,7 @@ module ROM
           # @return [Array<Hash>, Hash]
           #
           # @api private
-          def set_timestamps(tuples, *ignored)
+          def set_timestamps(tuples, *)
             timestamps = build_timestamps
 
             case tuples
--- a/lib/rom/sql/plugin/timestamps.rb
+++ b/lib/rom/sql/plugin/timestamps.rb
@@ -96,7 +96,7 @@ module ROM
           # @return [Array<Hash>, Hash]
           #
           # @api private
-          def set_timestamps(tuples)
+          def set_timestamps(tuples, *ignored)
             timestamps = build_timestamps
 
             case tuples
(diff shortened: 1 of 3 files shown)


lib/rom/sql/plugin/timestamps.rb
```diff
@@ -96,7 +96,7 @@ def datestamp_columns
           # @return [Array<Hash>, Hash]
           #
           # @api private
-          def set_timestamps(tuples)
+          def set_timestamps(tuples, *ignored)
```
solnic: `def set_timestamps(tuples, *)` should be enough

cflipse: updated

PR #166, opened 2017-03-01
Add :auto_restrictions plugin

### LINKED ISSUE #69: Plugin for generating common relations
How about adding a plugin that would generate relation views restricted by PK + any indexed column?

So ie:

``` ruby
class Users < ROM::Relation[:sql]
  schema do
    attribute :id, Types::Serial
    attribute :name, Types::Index(:int) # we don't have that yet btw
  end
end

# this would be generated automatically:
users.by_id
users.by_name
```

It's a common pattern so a bit of Convention-Over-Configuration here would be helpful.
##

solnic:
Closes #69

### DIFF
--- a/lib/rom/plugins/relation/sql/auto_restrictions.rb
+++ b/lib/rom/plugins/relation/sql/auto_restrictions.rb
@@ -1,0 +1,65 @@
+module ROM
+  module Plugins
+    module Relation
+      module SQL
+        # Generates methods for restricting relations by their indexed attributes
+        #
+        # This plugin must be enabled for the whole adapter, `use` won't work as
+        # schema is not yet available, unless it was defined explicitly.
+        #
+        # @example
+        #   rom = ROM.container(:sql, 'sqlite::memory') do |config|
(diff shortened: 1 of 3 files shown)


lib/rom/plugins/relation/sql/auto_restrictions.rb
```diff
@@ -0,0 +1,66 @@
...
+
+            indexed_attrs = klass.schema.select { |attr| attr.meta[:index] }
+
+            methods = indexed_attrs.map do |attr|
+              attr_name = attr.name
+              meth_name = :"by_#{attr_name}"
+
+              mod.module_eval do
+                define_method(meth_name) do |value|
+                  where(self.class.schema[attr_name].is(value))
```
flash-gordon: I wonder is it can be just `where(attr.is(value))`?

solnic: Oh very good point, it can totally use the attr we already have. I'll change this, thanks

PR #167, opened 2017-03-01
Fix grouping by a function

flash-gordon:
@solnic any clue why do we have that `.upcase` for the virtual row?


solnic:
Because reasons O_o

solnic:
@flash-gordon feel free to merge this in and even push a bug-fix release, this can easily bite people after all

flash-gordon:
aaaand it's done

solnic:
@flash-gordon fantastic 🎉

PR #171, opened 2017-03-06
Add support for joins by association name

flash-gordon:
Since association carries all info we need for a join it seems reasonable to infer join condition automatically.

### DIFF
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -811,7 +811,12 @@ module ROM
         def __join__(type, other, join_cond = EMPTY_HASH, opts = EMPTY_HASH, &block)
           case other
           when Symbol, Association::Name
-            new(dataset.__send__(type, other.to_sym, join_cond, opts, &block))
+            if join_cond.empty?
+              assoc = associations[other]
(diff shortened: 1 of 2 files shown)


flash-gordon:
It also can be sutin like `join(using: :assoc_name)`

solnic:
We already have `join(another_relation)` which does exactly that

flash-gordon:
Not exactly. You can have two associations with the same relation. And
relation join doesn't work if the association named differently from the
relation. I'll show an example later

flash-gordon:
@solnic I pushed an example https://github.com/rom-rb/rom-sql/pull/171/commits/131e927f27b2eb2eeff7eefc180200eaf7cdbfd0

solnic:
oh I see, this makes sense. I'm not a big fan of using symbols though. I also planned to make relations dumpable to sql literals, which is somewhat related, because we could have `join(users.as(:author))` and use its `alias` to construct required join. I'm OK with supporting plain symbols in this case, but eventually I'd like to support passing aliased relations too. WDYT?

flash-gordon:
interesting, yes, this makes a lot of sense. BTW I don't like symbols either. I'll mark the failing spec as pending

PR #179, opened 2017-04-12
Support for schema plugins

flash-gordon:
@solnic good news, `auto_restrictions` now can be enabled in any place.


flash-gordon:
ref https://github.com/rom-rb/rom/pull/389

flash-gordon:
@solnic updated and ready to go, though I don't like how `relation_schema_spec.rb` looks like now, but it's time to 💤 :)

lib/rom/plugins/relation/sql/auto_restrictions.rb
```diff
@@ -26,22 +23,31 @@ module SQL
...
             methods, mod = restriction_methods(schema)
             klass.include(mod)
             methods.each { |meth| klass.auto_curry(meth) }
           end
 
+          module FinalizationHook
+            def set_schema!(*)
+              super.tap { AutoRestrictions.apply(self) }
+            end
+          end
```
solnic: I will try to get rid of this in favor of setup/finalization events like we talked about. It's fine for now though, it shouldn't block further development, so I'm merging this regardless.

PR #184, opened 2017-04-19
Support for type extensions in Query and Projection DSLs

flash-gordon:
@solnic wdyt? If it looks fine overall I'll add more operations.

### DIFF
--- a/lib/rom/sql/attribute.rb
+++ b/lib/rom/sql/attribute.rb
@@ -14,6 +14,28 @@ module ROM
       # Error raised when an attribute cannot be qualified
       QualifyError = Class.new(StandardError)
 
+      TypeExtensions = {}
+
+      # Registers a set of operations supported for a specific type
+      #
+      # @example
+      #   ROM::SQL::Attribute.register_type(ROM::SQL::Types::PG::JSONB) do
+      #     def contains(type, keys)
+      #       Sequel::Postgres::JSONBOp.new(type.meta[:name]).contains(keys)
+      #     end
+      #   end
+      #
+      # @param [Dry::Types::Type] type Type
+      #
+      # @api public
+      def self.register_type(type, &block)
+        raise ArgumentError, "Type #{ type } already registered" if TypeExtensions.key?(type)
+        mod = Module.new(&block)
+        ctx = Object.new.extend(mod)
+        functions = mod.instance_methods.each_with_object({}) { |m, ms| ms[m] = ctx.method(m) }
+        TypeExtensions[type] = functions
+      end
+
       # Return a new attribute with an alias
       #
       # @example
@@ -235,6 +257,11 @@ module ROM
         @sql_expr ||= (meta[:sql_expr] || Sequel[to_sym])
       end
 
+      # @api private
+      def type_extensions
+        @type_extensions ||= TypeExtensions[unwrap_type(type)] || EMPTY_HASH
+      end
+
       # Delegate to sql expression if it responds to a given method
       #
       # @api private
@@ -243,6 +270,8 @@ module ROM
           __cmp__(meth, args[0])
         elsif sql_expr.respond_to?(meth)
           meta(sql_expr: sql_expr.__send__(meth, *args, &block))
+        elsif type_extensions.key?(meth)
+          type_extensions[meth].(type, *args, &block)
         else
           super
         end
@@ -262,6 +291,17 @@ module ROM
           end
 
         Sequel::SQL::BooleanExpression.new(op, self, value)
+      end
+
+      # Unwraps an optional type
+      #
+      # @api private
+      def unwrap_type(type)
+        if type.respond_to?(:left) && type.left == Types::Strict::Nil
+          type.right
+        else
+          type
+        end
       end
     end
   end
--- a/lib/rom/sql/attribute.rb
+++ b/lib/rom/sql/attribute.rb
@@ -14,27 +14,60 @@ module ROM
       # Error raised when an attribute cannot be qualified
       QualifyError = Class.new(StandardError)
 
-      TypeExtensions = {}
-
-      # Registers a set of operations supported for a specific type
-      #
-      # @example
-      #   ROM::SQL::Attribute.register_type(ROM::SQL::Types::PG::JSONB) do
-      #     def contains(type, keys)
-      #       Sequel::Postgres::JSONBOp.new(type.meta[:name]).contains(keys)
-      #     end
-      #   end
-      #
-      # @param [Dry::Types::Type] type Type
+      # Type-specific methods
       #
       # @api public
-      def self.register_type(type, &block)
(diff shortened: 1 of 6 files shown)


lib/rom/sql/attribute.rb
```diff
@@ -263,6 +292,17 @@ def __cmp__(op, other)
 
         Sequel::SQL::BooleanExpression.new(op, self, value)
       end
+
+      # Unwraps an optional type
+      #
+      # @api private
+      def unwrap_type(type)
+        if type.respond_to?(:left) && type.left == Types::Strict::Nil
```
solnic: this can stay for now but please add `TODO: add Type#optional? method to dry-types and clean this up`

lib/rom/sql/attribute.rb
```diff
@@ -14,6 +14,28 @@ class Attribute < ROM::Schema::Attribute
...
+      #
+      # @param [Dry::Types::Type] type Type
+      #
+      # @api public
+      def self.register_type(type, &block)
+        raise ArgumentError, "Type #{ type } already registered" if TypeExtensions.key?(type)
+        mod = Module.new(&block)
+        ctx = Object.new.extend(mod)
+        functions = mod.instance_methods.each_with_object({}) { |m, ms| ms[m] = ctx.method(m) }
+        TypeExtensions[type] = functions
```
solnic: This is a bit tricky so let me make sure I understand this. So, `register_type` can define new methods (functions), that always receive attribute's `type` as the first argument, and potentially more args and/or a block. We store them under `Attribute#type_extensions` and that's how we can use them later in `#method_missing`. Does this sound right?

flash-gordon: Yes, I started with
```ruby
register(type, custom_method: -> (type, *args) { ... })
```
but it looks ugly and I thought then why not just plain old `def`s?
I also think that accepting type as the first argument is more explicit and faster (just a guess) than `instance_exec` on `type`.

lib/rom/sql/attribute.rb
flash-gordon: sure, will do

lib/rom/sql/attribute.rb
solnic: I think it's smart, in a good way. After all we need extensions per type, not for all attribute objects. One concern is that we may have name clashes with sequel's sql expressions, that we delegate to. Maybe a simple check if a given method name doesn't clash would be a good idea.

flash-gordon: right, good point! 
BTW this way of extending is how ruby's dispatching works, but instead of having an object with methods we have a hash/list of methods directly. It's a bit more convenient because you have _only_ methods you defined, nothing leaks from `Object` and `Kernel`.

lib/rom/sql/attribute.rb
```diff
@@ -235,6 +257,11 @@ def sql_expr
         @sql_expr ||= (meta[:sql_expr] || Sequel[to_sym])
       end
 
+      # @api private
+      def type_extensions
+        @type_extensions ||= TypeExtensions[unwrap_type(type)] || EMPTY_HASH
+      end
```
solnic: Can't we just pass type extensions when schema is building its attributes? This would eliminate the need for `unwrap_type` run-time method as well as `type_extensions` with lazy-assigment.

flash-gordon: I thought about it, this is not gonna work, consider this

```ruby
schema do
  attribute :foo, Types::String.optional
end
```
The `attribute` method accepts an already unwrapped type, we have to unwrap it anyway

solnic: Hmm, don't you think we could do it [here](https://github.com/rom-rb/rom/blob/master/lib/rom/schema.rb#L89-L91)? ie:

``` ruby
def self.attributes(attributes, attr_class)
  attributes.map { |type| attr_class.new(type, extensions: TypeExtensions[type]) }
end
```

This would require `TypeExtenions#[]` to unwrap when needed.

flash-gordon: Right, this makes sense, I just wanted to show that we need to unwrap it _somewhere_ anyway. I'll move unwrapping to `TypeExtenions.[]`

solnic: I know we need to, but we can avoid doing that at run-time and we can keep Attribute API smaller (just an option with reader, instead of 2 methods and lazy assignment). Do you think it makes sense to extend Attribute with Options, or just add a second optional arg to its constructor? We may need options in attributes but it could slow things down, as we create a lot of new attribute objects at run-time.

flash-gordon: oh, I wasn't thinking about run-time attribute creating, could you show me where this happens? Things like `int::count` I suppose...

solnic: @flash-gordon yeah in the SQL DSLs, just look at methods that create new expressions, they all are used at run-time and return new attribute instances

flash-gordon:
@solnic I added `Initializer` to `Attribute`, now `extensions` are evaluated via a default proc. This means they will be preserved in DSL calls. Does it solve the performance concerns?

solnic:
@flash-gordon I think this will be fine

lib/rom/sql/attribute.rb
```diff
@@ -14,6 +14,49 @@ class Attribute < ROM::Schema::Attribute
...
+        class << self
+          # Gets extensions for a type
+          #
+          # @param [Dry::Types::Type] type
+          #
+          # @return [Hash]
+          #
+          # @api public
+          def [](type)
+            unwrapped = type.optional? ? type.right : type
```
solnic: I wonder if it's possible to have a case where `type.right` would be...another optional *heh*

lib/rom/sql/attribute.rb
```diff
@@ -14,6 +14,49 @@ class Attribute < ROM::Schema::Attribute
...
+            ctx = Object.new.extend(mod)
+            functions = mod.instance_methods.each_with_object({}) { |m, ms| ms[m] = ctx.method(m) }
+            @types[type] = functions
+          end
+        end
+
+        @types = {}
+      end
+
+      option :extensions, type: Dry::Types['hash'], default: -> { TypeExtensions[type] }
```
solnic: Please use `Types::Hash` (from `rom/types`)

lib/rom/sql/attribute.rb
flash-gordon: DB type cannot be nullable _twice_

solnic: yeah but...technically speaking somebody could do `attribute :foo, Types::String.optional | Types::Int.optional` for whatever reason (like using sqlite 😆). Just thinking that maybe adding a simple check and raising an arg error if `type.right` is also optional would reduce the risk of having awkward backtraces.

Anyhow, just thinking out loud, let's leave it and wait how it works out.

flash-gordon:
@solnic OK, it's finished

Issue #190, opened 2017-04-26
Negation attribute does not work


alexandru-calinoiu:
Give I have a very simple schema with a column `status` as string.

The following query does not seem to produce the required result:

```
where((!status.is('running')) || (status.is(nil))
```

The result  is a `WHERE ... AND 0`

Expected result is `WHERE ... AND status != 'running' OR status IS NULL`

AMHOL:
Hmm, perhaps we're just missing a `#not` method.

solnic:
This is done in master. It'll be available in 1.3.0.

PR #191, opened 2017-04-27
Add ROM::SQL::Attribute#not (refs #190)

AMHOL:


flash-gordon:
@AMHOL thank you! Have you tested it with `nil`? I wonder if `foo.not(nil)` transforms to `"foo" IS NOT NULL`

@solnic I see having `not` to be more clear than `!`, wdyt?

solnic:
@flash-gordon we should have both because people will intuitively try negation operator and it'd be awesome if that worked too

lib/rom/sql/attribute.rb
```diff
@@ -154,6 +154,20 @@ def is(other)
...
+      #   users.where { id.not(1) }
+      #
+      #   users.where(users[:id].not(1))
+      #
+      # @param [Object] other Any SQL-compatible object type
+      #
+      # @api public
+      def not(other)
+        __cmp__(:'!=', other)
+      end
```
solnic: You can make it more flexible by doing `~is(other)` which uses sequel's negation operator (not sure why it's not called `!` though)

flash-gordon:
call me biased but I'd not call it ever in a DSL :)

solnic:
@flash-gordon that's why in ruby we have `alias_method` so that we can all be happy :)

flash-gordon:
haha, that's true

AMHOL:
@solnic @flash-gordon should be good to go now

lib/rom/sql/attribute.rb
```diff
@@ -151,7 +152,25 @@ def to_sym
       #
       # @api public
       def is(other)
-        __cmp__(:'=', other)
+        if NONSTANDARD_EQUALITY_VALUES.include?(other)
+          __cmp__(:IS, other)
+        else
+          __cmp__(:'=', other)
+        end
+      end
```
solnic: Wait, I thought you're going to use `=~`? So:

``` ruby
def =~(o)
  meta(sql_expr: sql_expr =~ o)
end

def is(o)
  self =~ o
end
```

AMHOL: Ended up with failing specs because __cmp__ seems to handle passing types too: https://github.com/rom-rb/rom-sql/blob/master/lib/rom/sql/attribute.rb#L261

solnic:
@AMHOL @flash-gordon I made this stuff work with Sequel's `=~` and added `#!`. How does it look?

lib/rom/sql/attribute.rb
```diff
@@ -11,6 +11,7 @@ module SQL
     # @api public
     class Attribute < ROM::Schema::Attribute
       OPERATORS = %i[>= <= > <].freeze
+      NONSTANDARD_EQUALITY_VALUES = [true, false, nil].freeze
```
AMHOL: Don't think this is used now?

AMHOL:
Nice one @solnic :smile:

PR #197, opened 2017-05-03
Fix Sequel warnings

flash-gordon:
@solnic this requires a thorough look. We definitely have a leaking abstraction regarding qualified SQL names

### DIFF
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -105,7 +105,7 @@ module ROM
         #
         # @api public
         def qualified_columns
-          schema.qualified.map(&:to_sym)
+          schema.qualified.map(&:to_sql_name)
         end
 
         # Map tuples from the relation
@@ -341,8 +341,10 @@ module ROM
             where(*args).where(self.class.schema.restriction(&block))
           elsif args.size == 1 && args[0].is_a?(Hash)
             new(dataset.where(coerce_conditions(args[0])))
-          else
+          elsif !args.empty?
             new(dataset.where(*args))
+          else
+            self
           end
         end
 
@@ -395,9 +397,9 @@ module ROM
         # @api public
         def having(*args, &block)
           if block
-            new(dataset.having(*args).having(self.class.schema.restriction(&block)))
+            new(dataset.having(*args, *self.class.schema.restriction(&block)))
           else
-            new(dataset.__send__(__method__, *args, &block))
+            new(dataset.__send__(__method__, *args))
           end
         end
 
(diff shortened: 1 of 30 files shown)


lib/rom/sql/relation/reading.rb
```diff
@@ -341,8 +341,10 @@ def where(*args, &block)
             where(*args).where(self.class.schema.restriction(&block))
           elsif args.size == 1 && args[0].is_a?(Hash)
             new(dataset.where(coerce_conditions(args[0])))
-          else
+          elsif !args.empty?
             new(dataset.where(*args))
+          else
+            self
           end
```
solnic: hmm, so, now you can call `relation.where` and it just returns self?

solnic:
Do I understand correctly that this will make `where(foo__bar: 'baz')` stop working?

lib/rom/sql/relation/reading.rb
flash-gordon: ah, `Sequel` doesn't like it anymore, `new(dataset)` is another option here

solnic: doesn't like what exactly?

flash-gordon:
@solnic it works, just issues a warning. I think I'll just set `split_symbols = true` for some specs so that we can ensure it works.

lib/rom/sql/relation/reading.rb
flash-gordon: calling `where/having` without args and a block

flash-gordon:
```
SEQUEL DEPRECATION WARNING: Passing no arguments and no block to a filtering method is deprecated and will be removed in Sequel 5.  Include at least one argument or a block when calling a filtering method.
```

solnic:
@flash-gordon yeah it'd be good, we'll get rid of this in 2.0 anyway

flash-gordon:
sure, will do, I forgot to add it's WIP 😆

lib/rom/sql/relation/reading.rb
solnic: OK so when do we do this?

flash-gordon: here https://github.com/rom-rb/rom-sql/blob/c426a83c65b701c8a7eb29052f5c23e29a54e446/lib/rom/sql/relation/reading.rb#L775-L777

flash-gordon:
@solnic works on my app, do we need anything else?

flash-gordon:
I'm merging this in master since I released v1.3.1

PR #199, opened 2017-05-12
Add PG::Array operations

flash-gordon:


solnic:
Fantastic :) 👍

flash-gordon:
😄

flash-gordon:
@solnic btw there is a problem with docs, there is a clash in method names, i.e. `Attribute<PG::JSONB>` and `Attribute<PG::Array>` both have `contain` and only one gets into the docs. Perhaps I could use `overloaded` for this... will give it a shot tonight

solnic:
@flash-gordon yes you should be able to use `@overload` to fix this

flash-gordon:
@solnic hmm, this works but I have to put all docs in one place, in a giant docstring with macros, that's pretty crazy :) I'll push it later

flash-gordon:
meh, I'm not happy with the result, I'll release it as is and will look into creating a plugin for yard

solnic:
@flash-gordon you could ask [Loren](https://github.com/lsegal) if he has any suggestions

flash-gordon:
@solnic I think what we need here is kind of docs for generics, i.e. list of methods for `ROM::SQL::Attribute<PG::Array>`, `ROM::SQL::Attribute<PG::JSONB>`, etc

PR #200, opened 2017-05-14
Support for row-level locks

flash-gordon:
This at least adds support for PosqtreSQL, MySQL, and Oracle. SQLite doesn't have row-level locks.

I will try it in my app first, but it's more or less complete.

### DIFF
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -7,6 +7,22 @@ module ROM
       #
       # @api public
       module Reading
+        # Row-level lock modes
+        ROW_LOCK_MODES = Hash.new(update: 'FOR UPDATE'.freeze).update(
+          # https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FOR-UPDATE-SHARE
+          postgres: {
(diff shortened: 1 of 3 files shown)


backus:
Nice! The current implementation we have looks like this

```ruby
# frozen_string_literal: true

class Posts < ROM::Relation[:sql]
  schema(:posts) do
    # ...
  end

  def lock_for_update
    new(dataset.lock_style('FOR UPDATE NOWAIT'))
  end
end # Posts
```

Am I correct in assuming that we would replace our calls to `lock_for_update` with `lock(wait: false)`?

flash-gordon:
@backus right, three common uses
```ruby
lock # => FOR UPDATE
lock(wait: false) # => FOR UPDATE NOWAIT
lock(skip_locked: true) # => FOR UPDATE SKIP LOCKED
```

backus:
Looks great! Nice job. I'm impressed you remembered this feature. I don't think I even opened an issue

flash-gordon:
It's not that surprising tbh, I'm rewriting code that uses AR's locks :)

solnic:
Great job! LGTM

PR #201, opened 2017-05-15
Add support for using CAST for converting data types

flash-gordon:


lib/rom/sql/function.rb
```diff
@@ -26,6 +26,21 @@ def is(other)
         ::Sequel::SQL::BooleanExpression.new(:'=', func, other)
       end
 
+      # Convert an expression result to another data type
+      #
+      # @example
+      #   users.select { bool::cast(json_data.get_text('activated'), :boolean).as(:activated) }
```
solnic: We should be able to infer cast types based on type spec, so this could be simplified to be:

``` ruby
users.select { bool::json_data.get_text('activated').as(:activated) }
```

WDYT?

flash-gordon: you're absolutely right, but this will require a DB-specific mapping between data types (e.g. oracle uses `varchar2` instead of `varchar`), that's why I decided to make it explicit for now, I'll open a help-requested issue for this improvement :)

solnic: Yes starting with an explicit API is definitely a great first step!

flash-gordon:
I must admit this works surprisingly well 🎉

lib/rom/sql/function.rb
```diff
@@ -26,6 +26,21 @@ def is(other)
...
+      #
+      # @param [ROM::SQL::Attrbite] expr Expression to be cast
```
solnic: `Attrbite` => `Attribute` (I know, too late ;))

PR #203, opened 2017-05-22
Add SQL `EXISTS` expression to Restriction DSL

v-kolesnikov:


flash-gordon:
awesome 👍

v-kolesnikov:
👌

solnic:
Would it make sense to have `Relation#exists` so that you can simply do `users.where(other.exists)`?

flash-gordon:
@solnic I think so, we need a convenient way to correlate queries/subqueries other, with the block syntax we could do something like this
```ruby
users.order(:name).where { exists(tasks, tasks[:user_id].qualified => id.qualified) }
```
So in case of `Relation#exists` this can be
```ruby
users.where(other.exists(other[:user_id].qualified => users[:id].qualified))
```

solnic:
@flash-gordon but `users.where(other.where(...).exists))` is doable too and should be way more flexible (I think)

flash-gordon:
@solnic both options should be allowed but I don't know any case where you want to use an uncorrelated `EXISTS` so I'd like people to know that they usually need to pass _something_ into `exists`

solnic:
@flash-gordon IIRC it's possible to determine if a relation was restricted, so if I'm right, then we can raise an error if somebody tries to do `exists` on not-restricted relation. WDYT?

flash-gordon:
@solnic good point, also shouldn't it be `relation.exists(other)`? I don't like the hanging `exists` call

solnic:
@flash-gordon so the assumption would be that anything that's passed to `exists` would delegate to `where` and then return a new relation with `dataset.exists`?

flash-gordon:
@solnic right, both `relation.exists(other)`, `relation.exists(other, join => condition)` should be allowed, the latter form makes more sense because I wrote a lot of SQL code and didn't see a single case where `EXISTS` wasn't correlated with the main query

solnic:
OK sounds good

PR #209, opened 2017-06-17
Adding API for auto-migrations

flash-gordon:
This adds support for automatic DB schema migrations based on relation schemas. Target API will look similar to a single simple call `gateway.auto_migrate!(container)`

- [x] Creating a table
- [x] Adding columns to an existing table
- [ ] Renaming columns (without coercion)
- [x] Removing columns
- [ ] Support for default types
- [ ] Emitting migration files
- [x] Friendly errors on unsupported operations
- [x] Managing indexes
- [x] Managing nullability
- [ ] Managing FK constraints

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,7 +9,7 @@ gem 'rom', git: 'https://github.com/rom-rb/rom', branch: 'master' do
 end
 
 group :test do
-  gem 'byebug', platforms: :mri
+  gem 'pry-byebug', platforms: :mri
   gem 'pry', platforms: %i(jruby rbx)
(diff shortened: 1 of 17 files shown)


flash-gordon:
I don't feel like we have to support renaming columns OOTB but we can go with something like

```ruby
attribute :full_name, Types::String.meta(from: :name)
```
This `:from` key is a temporary piece that needs for automatic migration only. After the migration has been generated it can be safely removed. @solnic wdyt? 

Overall, I'd like this to be merged in master because further changes require touching other stuff like `Schema`, `Attribute` and so on and I'd split it into parts.

PR #214, opened 2017-06-27
Add DSL for defining indexes

flash-gordon:
- [x] Add basic DSL
- [x] Support for arbitrary names
- [x] Add multi-column auto-restrictions (crazy)
- [x] Unique indexes
- [x] Different index types (not sure we can manage them later)
- [x] Partial indexes in PG (?)
- [ ] Finally add PK to the index list (?)

That should be a good start :)

### DIFF
--- a/lib/rom/plugins/relation/sql/auto_restrictions.rb
+++ b/lib/rom/plugins/relation/sql/auto_restrictions.rb
@@ -39,18 +39,29 @@ module ROM
           def self.restriction_methods(schema)
             mod = Module.new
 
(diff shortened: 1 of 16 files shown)


solnic:
O M G 😍

flash-gordon:
I don't feel like adding PK to the index list would help us at this point so we'll return to this later

flash-gordon:
For the record, we don't support index type switching, we even can't infer index type because Sequel doesn't return it. It's not a big problem FWIW, we'll figure out this later, _somehow_

PR #216, opened 2017-07-02
Managing foreign keys in auto-migrations

flash-gordon:
This adds support for managing foreign key constraints within a single gateway. Foreign key list that I added is kind of generic and can be moved later to the core

- [x] Add list of foreign key to the schema
- [x] Check if FKs work fine with the inferrer
- [x] Manage FKs in auto-migrations. It must be the last step because FKs are safe to be created only when all tables are in place

### DIFF
--- a/lib/rom/sql/types.rb
+++ b/lib/rom/sql/types.rb
@@ -14,7 +14,7 @@ module ROM
       end
 
       def self.ForeignKey(relation, type = Types::Int.meta(index: true))
-        type.meta(foreign_key: true, target: relation)
+        super
       end
 
       Serial = Int.meta(primary_key: true)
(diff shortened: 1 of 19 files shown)


lib/rom/sql/types.rb
```diff
@@ -13,6 +13,10 @@ def self.Definition(*args, &block)
         ROM::Types.Definition(*args, &block)
       end
 
+      def self.ForeignKey(relation, type = Types::Int.meta(index: true))
+        type.meta(foreign_key: true, target: relation)
+      end
```
solnic: Shouldn't this be:

``` ruby
def self.ForeignKey(relation, type = Types::Int)
  type.meta(foreign_key: true, index: true, target: relation)
end
```

??

flash-gordon: hm, right

PR #218, opened 2017-07-09
Create migration files in auto-migrations

flash-gordon:
- [x] Create table
- [x] Alter table
- [x] Test various options for tables/indexes/FKs

### DIFF
--- a/lib/rom/sql/migration.rb
+++ b/lib/rom/sql/migration.rb
@@ -138,12 +138,12 @@ module ROM
       end
 
       # @api public
-      def auto_migrate!(conf)
+      def auto_migrate!(conf, options = EMPTY_HASH, &block)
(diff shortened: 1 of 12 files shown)


flash-gordon:
@solnic the most difficult part so far is naming 😥

solnic:
@flash-gordon naming what? :) Don't cry 🤗

flash-gordon:
specifically those runner/inline_runner/writer/whatever. Naming them with verbs would be easier

solnic:
So name them with verbs :) WDYT about using template files instead of generating stuff by concatenating strings?

flash-gordon:
@solnic I wouldn't bother at this point, this stuff takes ~30 lines and it not going to get any bigger, we'll see though

PR #219, opened 2017-07-10
Qualify schemas by default

solnic:


flash-gordon:
@solnic I force pushed the branch

PR #225, opened 2017-07-13
Add SQL::Relation#import

flash-gordon:
This method import data from another relation. In common cases, it will use the `INSERT ... SELECT ...` statement which is known as being very effective.

### DIFF
--- a/lib/rom/sql/relation/writing.rb
+++ b/lib/rom/sql/relation/writing.rb
@@ -77,6 +77,44 @@ module ROM
         def delete(*args, &block)
           dataset.delete(*args, &block)
         end
+
+        # Insert tuples from other relation
+        # NOTE: The method implicitly uses a transaction
+        #
+        # @example
+        #   users.import(new_users)
+        #
+        # @overload import(other_sql_relation, options)
+        #   If both relations uses the same gateway
+        #   the INSERT ... SELECT statement will
+        #   be used for importing the data
+        #
+        #   @params [SQL::Relation] other_sql_relation
+        #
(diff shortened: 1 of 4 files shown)


lib/rom/sql/relation/writing.rb
```diff
@@ -77,6 +77,44 @@ def update(*args, &block)
         def delete(*args, &block)
           dataset.delete(*args, &block)
         end
+
+        # Insert tuples from other relation
+        # NOTE: The method is implicitly uses a transaction
```
cflipse: ~is~

lib/rom/sql/relation/writing.rb
```diff
@@ -77,6 +77,44 @@ def update(*args, &block)
...
+        #
+        # @example
+        #   users.insert_from(new_users)
```
cflipse: should update to import here.

lib/rom/sql/relation/writing.rb
```diff
@@ -77,6 +77,44 @@ def update(*args, &block)
...
+        #   @option [Integer] :slice
+        #     Split loading into batches of provided size,
+        #     every batch will be processed in a separate
+        #     transaction block
+        #
+        # @overload import(other, options)
+        #   Import data from another relation. The source
+        #   relation will be materialized before loading
+        #
+        #   @params [Relation] other_sql_relation
```
cflipse: is other necessarily an SQL relation here?

flash-gordon:
@cflipse thanks for the review ❤️

PR #226, opened 2017-07-13
Discard old rows before creating new in 1-to-1 associations

flash-gordon:
@solnic please review. I decided not to wrap it in a transaction block because in my opinion this should be done on a higher level. We already have one-to-many (and many-to-many) commands that don't use transactions so I suggest being consistent :)

/cc @mereghost

### DIFF
--- a/lib/rom/sql/association/one_to_one.rb
+++ b/lib/rom/sql/association/one_to_one.rb
@@ -3,6 +3,13 @@ module ROM
     class Association
       class OneToOne < OneToMany
         result :one
+
+        # @api private
+        def remove_associated(relations, parent)
+          pk, fk = join_key_map(relations)
+          relation = relations[source.relation]
+          relation.where(fk => parent.fetch(pk)).delete
+        end
       end
     end
   end
--- a/lib/rom/sql/plugin/associates.rb
+++ b/lib/rom/sql/plugin/associates.rb
@@ -154,6 +154,11 @@ module ROM
                 else
                   tuples.map { |tuple| Hash(tuple).update(fk => parent[pk]) }
                 end
+              when Association::OneToOne
+                with_input_tuples(tuples).map { |tuple|
+                  assoc.remove_associated(__registry__, parent)
+                  assoc.associate(__registry__, tuple, parent)
+                }
               when Association
                 with_input_tuples(tuples).map { |tuple|
                   assoc.associate(__registry__, tuple, parent)
--- a/spec/integration/plugins/associates_spec.rb
+++ b/spec/integration/plugins/associates_spec.rb
@@ -192,6 +192,71 @@ RSpec.describe 'Plugins / :associates', seeds: false do
             expect(tags).to eql(result)
           end
         end
+
+        context 'with one-to-one association' do
+          before do
+            conn.drop_table?(:profiles)
+
+            conn.create_table :profiles do
+              primary_key :id
+              column :user_id, Integer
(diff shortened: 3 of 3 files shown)


mereghost:
LGTM

solnic:
I think this is fine, at least for a good start.

AMHOL:
Do we want an option to just unlink the association rather than delete it?

flash-gordon:
@AMHOL in a typical model this would end up with a not-null constraint error. It could be an option but I don't see how to configure it, mb @solnic has some thoughts on this

solnic:
We could look if the fk has not-null constraint or not, and if it has, then we delete, if it doesn't, we set it to nil. I suppose.

flash-gordon:
It's an easy thing to do in rom but this makes the whole algorithm even more hard to understand fwiw. That's the reason I don't like ActiveRecord

solnic:
So, should we merge this in?

flash-gordon:
yeah, I'll do this... I mean I did, but for some reason, the PR didn't get close automatically. Anyhow, I need to add this to the master branch

mereghost:
Awesome news. =D Waiting for the release.

flash-gordon:
Folks, I got a question regarding this. Let's consider an example, a user has one profile. There are tables `users` and `profiles` and `profiles` references `users` with `user_id` as a FK. Now a user `has_one` profile, but a profile `belongs_to` a user. This means you can create a user with a profile using the `has_one` association, not vice versa. Obviously, a new user cannot have a profile yet so we don't need to delete anything at all. As a result, we cannot maintain the `has_one` guarantee with ROM because on the profile's side we have a `belongs_to` association (and we haven't a concept of the `inverse_of` option from AR).

Could anyone correct me?

P.S. Please don't ask me why I didn't say it before, I just use this type of assoc very occasionally :)

mereghost:
Does @solnic suggestion of checking for the not null constraint solves this case, @flash-gordon ?

jodosha:
@flash-gordon 
> Obviously, a new user cannot have a profile yet so we don't need to delete anything at all.

Can a semantic like "drop if exist" be applied here?

flash-gordon:
@jodosha @mereghost this is what I tried to implement initially but this is not going to work, I'll try to break it down here. `hanami-model` uses associations like this

```ruby
def associate(data)
  relation(source)
    .associations[target]
    .associate(container.relations, data, subject)
end

```

And here is `associate` from rom-sql
```ruby
# @api private
def associate(relations, child, parent)
  pk, fk = join_key_map(relations)
  child.merge(fk => parent.fetch(pk))
end
```

As you can see, it is about updating hashes, not about issuing DML statements. It is also a private API and has been moved to the `rom-core` gem in `rom` 4.0 (hence can't access SQL now).

What ROM has as a public API is associating commands, i.e.
```ruby
user_repo.command(:create, user_repo.aggregate(:profile)).call(
  name: 'Jane Doe',
  profile: {
    picture: 'http://...'
  }
)
```
^this creates a user with theirs profile in one go but you simply cannot create a user _twice_ thus you don't need to drop anything :)

It looks to me we don't have an API to be "fixed" :) Hanami could remove existing records in the `add` method

```ruby
def add(data)
  # remove existing records
  command(:create, relation(target), use: [:timestamps]).call(associate(data))
end
```

@solnic did you plan to have an API for this?

solnic:
Going to close this since it's not going to be merged in after all.

PR #229, opened 2017-07-27
Add Relation#each_batch

flash-gordon:
This method works similar to ActiveRecord's Model.in_batches. From time to time,
it's useful to process a big dataset in a few steps.

### DIFF
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -880,6 +880,45 @@ module ROM
           where(other.where(join_condition).dataset.exists)
         end
 
+        # Process the dataset in batches.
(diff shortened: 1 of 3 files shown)


lib/rom/sql/relation/reading.rb
```diff
@@ -880,6 +880,46 @@ def exists(other, condition = nil)
...
+        #
+        # @api public
+        def each_batch(size: 1000)
+          pks = schema.primary_key
+
+          if pks.size > 1
+            raise ArgumentError, 'Composite primary keys are not supported yet'
+          end
+
+          pk_name = schema.primary_key_names[0]
```
solnic: There's a shortcut for this `Relation#primary_key`

PR #230, opened 2017-07-31
Add missing support for *text types in MySQL

### LINKED ISSUE #183: Can't infer columns with type LONGTEXT
I checked Grand Comics Database dump (mysql version, because postgresql compatible version doesn't work right now) with rom-sql. Unfortunatelly it doesn't work when table has LONGTEXT datatype column. I had to write:

```ruby
  schema(infer: true) do
    attribute :notes, Types::String # LONGTEXT
  end
```

After this, for notes I am receiving:
```
#<Sequel::SQL::Blob:0x3fe7bb011df4 bytes=162 start="Commonly c" end="0s titles.">
```

panthomakos:
In MySQL it's possible to define `tinytext`, `text`, `mediumtext` or
`longtext` column types. `rom-sql` currently only supports `text`
and `mediumtext` attributes. This change ensures that `tinytext` and
`longtext` are also supported and adds tests to validate the behavior
for all text types.

Fixes #183

### DIFF
--- a/lib/rom/sql/extensions/mysql/type_builder.rb
+++ b/lib/rom/sql/extensions/mysql/type_builder.rb
@@ -4,6 +4,22 @@ module ROM
   module SQL
     module MySQL
       class TypeBuilder < Schema::TypeBuilder
+        defines :db_type_mapping
+
+        db_type_mapping(
+          'tinytext' => Types::String,
+          'text' => Types::String,
+          'mediumtext' => Types::String,
+          'longtext' => Types::String
+        ).freeze
+
+        def map_type(ruby_type, db_type, **_)
+          map_db_type(db_type) || super
+        end
+
+        def map_db_type(db_type)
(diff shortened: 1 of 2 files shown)


flash-gordon:
Thanks! Since these text types are specific to MySQL (that is are not listed [here](https://www.w3schools.com/sql/sql_datatypes_general.asp)) could you move them to a separate type builder? In the same way this [was done](https://github.com/rom-rb/rom-sql/blob/40cf9183e90992ba6c80860cb0f71cc6e24f9eed/lib/rom/sql/extensions/postgres/type_builder.rb) for PostgreSQL. Also, the specs should be put in [mysql_spec.rb](https://github.com/rom-rb/rom-sql/blob/40cf9183e90992ba6c80860cb0f71cc6e24f9eed/spec/integration/schema/inferrer/mysql_spec.rb)

flash-gordon:
Great! Thank you! ❤️

PR #237, opened 2017-08-21
Combine with natural key

cflipse:
Adding a couple of failing specs for some scenarios I'm hitting in the wild.

The first set is a reproduction of a bug that I'm actively seeing.  The second is chasing after syntax for a join; I can remove that commit from this if necessary to keep things focused.


cflipse:
As these are repro bugs, the failing specs are expected; it's just easiest to get commentary using a PR

solnic:
I should mention that with assocs configured with custom combine keys and overridden views, you can just do `users.combine(:drafts)`.

Issue #236, opened 2017-08-18
Support for many_through_many associations


dikond:
Hi! This issue was extracted from the gitter discussion, here is [the link](https://gitter.im/rom-rb/chat?at=5996a4b980d90ca0240fe4b5).

The following example will tell better than I will.
```ruby
class Eans < ROM::Relation[:sql]
  schema(:eans, infer: true) do
    associations do
      has_many :ean_stats
      has_many :contract_ean_stats, through: :ean_stats
      has_many :contracts,          through: :contract_ean_stats
      has_many :companies,          through: :contracts
    end
  end
end

eans.combine(:contract_ean_stats).limit(1).one
# => ROM::Struct::Ean
eans.combine(:contracts).limit(1).one
# => KeyError: :ean_id attribute doesn't exist in contract_ean_stats schema
```

dikond:
Here is a self-contained snippet
```ruby
require 'rom'
require 'sqlite3'

rom = ROM.container(:sql, 'sqlite::memory') do |conf|
  conf.default.create_table(:eans) do
    primary_key :id
  end

  conf.default.create_table(:ean_stats) do
    primary_key :id
    column :ean_id, Integer
  end

  conf.default.create_table(:contract_ean_stats) do
    primary_key :id
    column :contract_id, Integer
    column :ean_stat_id, Integer
  end

  conf.default.create_table(:contracts) do
    primary_key :id
  end

  conf.relation(:eans) do
    schema(infer: true) do
      associations do
        has_many :ean_stats
        has_many :contract_ean_stats, through: :ean_stats
        has_many :contracts,          through: :contract_ean_stats
      end
    end
  end

  conf.relation(:ean_stats) do
    schema(infer: true) do
      associations do
        belongs_to :ean
        has_many :contract_ean_stats
      end
    end
  end

  conf.relation(:contract_ean_stats) do
    schema(infer: true) do
      associations do
        belongs_to :contract
        belongs_to :ean_stat
      end
    end
  end

  conf.relation(:contracts) do
    schema(infer: true) do
      associations do
        has_many :contract_ean_stats
        has_many :ean_stats, through: :contract_ean_stats
      end
    end
  end
end

rom.relations[:eans].combine(:contract_ean_stats).one
# => nil
rom.relations[:eans].combine(:contracts).one
# => KeyError: :ean_id attribute doesn't exist in contract_ean_stats schema
```

solnic:
[...]

PR #238, opened 2017-08-23
Many-to-Many scenario with no real FKs done via overridden view

solnic:
refs #236

### DIFF
--- a/spec/integration/associations/many_to_many/multiple_spec.rb
+++ b/spec/integration/associations/many_to_many/multiple_spec.rb
@@ -65,8 +65,9 @@ RSpec.describe ROM::SQL::Associations::ManyToMany, helpers: true do
 
         def for_eans(assoc, eans)
           join(:contract_ean_stats, contract_id: contracts[:id]).
-            join(:ean_stats, ean_id: eans.pluck(:id)).
-            select_append(ean_stats[:ean_id])
+            join(:ean_stats, id: contract_ean_stats[:ean_stat_id]).
+            select_append(ean_stats[:ean_id]).
+            where(ean_id: eans.pluck(:id))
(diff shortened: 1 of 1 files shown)


spec/integration/associations/many_to_many/multiple_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+        schema(infer: true) do
+          associations do
+            has_many :contract_ean_stats
+            has_many :ean_stats, through: :contract_ean_stats
+          end
+        end
+
+        def for_eans(assoc, eans)
+          join(:contract_ean_stats, contract_id: contracts[:id]).
+            join(:ean_stats, ean_id: eans.pluck(:id)).
```
flash-gordon: @solnic if Sequel doesn't do anything I don't know about, you did a cartesian join here and, usually, you don't need such a join :)

this should be similar to
```ruby
join(:contract_ean_stats, contract_id: contracts[:id]).
  join(:ean_stats, contract_id: contracts[:id]).
  select_append(ean_stats[:ean_id]).
  where(ean_id: eans.pluck(:id))
```

solnic: @flash-gordon yeah I had a feeling something ain't right. I'd actually appreciate if you could pull in this branch locally and see if this works. IIRC I tried what you wrote here but it didn't work (query was invalid)

flash-gordon: @solnic agh, right, I was wrong. Fixed it here https://github.com/rom-rb/rom-sql/pull/238/commits/91d4d755b48d583bdc8e935afa101abcd9dc88b3

PR #242, opened 2017-09-25
Add support for the FILTER clause

flash-gordon:
🎉

### DIFF
--- a/lib/rom/sql/function.rb
+++ b/lib/rom/sql/function.rb
@@ -132,7 +132,38 @@ module ROM
         Attribute[type].meta(sql_expr: ::Sequel.cast(expr, db_type))
       end
 
+      # Add a FILTER clause to aggregate function (supported by PostgreSQL 9.4+)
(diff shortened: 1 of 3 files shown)


flash-gordon:
Say you want to get a number of users who registered less than a day ago, a week ago and a month ago. All this using a single query. With FILTER it's almost trivial

```ruby
users.select {
  [int::count(:id).filter { created_at > time::now() - any::cast('1 day', 'interval') }.as(:last_day_users),
   int::count(:id).filter { created_at > time::now() - any::cast('1 week', 'interval') }.as(:last_week_users),
   int::count(:id).filter { created_at > time::now() - any::cast('1 month', 'interval') }.as(:last_month_users)]
}.order(nil)
```

solnic:
Would be good to move it to a PG-specific extension. WDYT?

flash-gordon:
@solnic technically, it’s a sql’s standard, it’s just only PG supports it right now

solnic:
I see, OK merging in.

Issue #234, opened 2017-08-09
[inferrer] Support for PG range types


flash-gordon:
I personally haven't had a chance to work with range types so one should read the docs before working on this issue: https://www.postgresql.org/docs/current/static/rangetypes.html 😄

solnic:
I did. I even took a stab at adding support for them to rom-sql but gave up because it wasn't a half-an-hour-task :)

solnic:
Done via #245 🎉

PR #245, opened 2017-10-01
Add PostgreSQL range types support

v-kolesnikov:
Hi, all!

I tried to add Postgres range types support described in #234 
Please comment as much as need it.

TO-DO:
- [x] Fix failure on the test build with JRuby
- [x] Add range [operators](https://postgrespro.com/docs/postgresql/9.6/functions-range.html#range-operators-table)
- [x] Add tests for range functions
- [x] Add cache for range parsers


lib/rom/sql/extensions/postgres/types/range.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      module Types
+        def self.Range(name, subtype)
+          Type(name) do
+            read = SQL::Types.Constructor(Values::Range) do |value|
+              pg_range =
+                if value.is_a?(Sequel::Postgres::PGRange)
+                  value
+                elsif value.respond_to?(:to_s)
+                  Sequel::Postgres::PGRange::Parser
```
flash-gordon: since parsers are like pure functions we should cache them somewhere, a global cache would be fine

lib/rom/sql/extensions/postgres/types/range.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+
+      module Types
+        def self.Range(name, subtype)
+          Type(name) do
+            read = SQL::Types.Constructor(Values::Range) do |value|
+              pg_range =
+                if value.is_a?(Sequel::Postgres::PGRange)
+                  value
+                elsif value.respond_to?(:to_s)
```
flash-gordon: `nil` responds to `to_s` but when we see `nil` we should return it back

v-kolesnikov:
I checked other PG libraries ([Postgrex](https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/builtins.ex#L19-L38), [Sequel](https://github.com/jeremyevans/sequel/blob/master/lib/sequel/extensions/pg_range.rb#L333-L338)) and now I think that this format for Range type values will be better:

```ruby
module Values
  Range = Struct.new(:lower, :upper, :lower_inclusive, :upper_inclusive)
end
```

WDYT?

lib/rom/sql/extensions/postgres/types/range.rb
```diff
@@ -6,7 +6,19 @@ module ROM
...
+            super
+          end
+
+          def exclude_begin?
+            bounds[0] == '('
+          end
+
+          def exclude_end?
+            bounds[1] == ')'
+          end
```
solnic: It would be nice to avoid allocation of these strings and use constants instead, like `PAREN_LEFT = '('.freeze` and `PARENT_RIGHT = ')'.freeze`.

lib/rom/sql/extensions/postgres/types/range.rb
```diff
@@ -7,16 +7,19 @@ module SQL
...
           end
 
           def exclude_begin?
-            bounds[0] == '('
+            bounds[0] == PARENT_LEFT
           end
 
           def exclude_end?
-            bounds[1] == ')'
+            bounds[1] == PARENT_RIGHT
```
solnic: `PARENT_*` => `PAREN_*` :)

v-kolesnikov: Oh, sorry.

lib/rom/sql/extensions/postgres/types/range.rb
```diff
@@ -0,0 +1,223 @@
...
+
+        Int4Range = range('int4range', range_read_type(:int4range))
+        Int8Range = range('int8range', range_read_type(:int8range))
+        NumRange  = range('numrange',  range_read_type(:numrange))
+        TsRange   = range('tsrange',   range_read_type(:tsrange))
+        TsTzRange = range('tstzrange', range_read_type(:tstzrange))
+        DateRange = range('daterange', range_read_type(:daterange))
+
+        module RangeOperators
+          def contains(_type, expr, value)
```
flash-gordon: I think we should rename it to `contain`, it’s a semi-consistent ruby’s convention, as in `respond_to?`, `start_with?`, `include?` etc follow this.

lib/rom/sql/extensions/postgres/types/range.rb
```diff
@@ -0,0 +1,223 @@
...
+            )
+          end
+
+          def upper(_type, expr)
+            Attribute[SQL::Types::Bool].meta(
+              sql_expr: Sequel.pg_range(expr).upper
+            )
+          end
+
+          def isempty(_type, expr)
```
flash-gordon: Should be is_empty

lib/rom/sql/extensions/postgres/types/range.rb
```diff
@@ -0,0 +1,223 @@
...
+          end
+
+          def upper_inf(_type, expr)
+            Attribute[SQL::Types::Bool].meta(
+              sql_expr: Sequel.pg_range(expr).upper_inf
+            )
+          end
+        end
+
+        TypeExtensions.register(Int4Range) do
```
flash-gordon: Let’s replace the following registrations with a loop :)

PR #244, opened 2017-09-30
[PG] Add ltree support

GustavoCaso:
#145 

Add support for `ltree`

@flash-gordon is my first time adding a new Type please comment as much as need it, I would apply the changes happily

PR #246, opened 2017-10-02
Add ltree query methods

GustavoCaso:
Operator | Returns | Description | Done
-- | -- | -- | --
ltree @> ltree | boolean | is left argument an ancestor of right (or equal)? | X
ltree <@ ltree | boolean | is left argument a descendant of right (or equal)? | X
ltree ~ lquery | boolean | does ltree match lquery? | X
lquery ~ ltree | boolean | does ltree match lquery? | X
ltree ? lquery[] | boolean | does ltree match any lquery in array? | X
lquery[] ? ltree | boolean | does ltree match any lquery in array? | X
ltree @ ltxtquery | boolean | does ltree match ltxtquery? | X
ltree \|\| ltree | ltree | concatenate ltree paths | X
ltree \|\| text | ltree | convert text to ltree and concatenate | X
ltree[] @> ltree | boolean | does array contain an ancestor of ltree? | X
ltree <@ ltree[] | boolean | does array contain an ancestor of ltree? | X
ltree[] <@ ltree | boolean | does array contain a descendant of ltree? | X
ltree @> ltree[] | boolean | does array contain a descendant of ltree? | X
ltree[] ~ lquery | boolean | does array contain any path matching lquery? | X
ltree[] ? lquery[] | boolean | does ltree array contain any path matching any lquery? | X
ltree[] @ ltxtquery | boolean | does array contain any path matching ltxtquery? | X
ltree[] ?@> ltree | ltree | first array entry that is an ancestor of ltree; NULL if none | X
ltree[] ?<@ ltree | ltree | first array entry that is a descendant of ltree; NULL if none | X
ltree[] ?~ lquery | ltree | first array entry that matches lquery; NULL if none | X
ltree[] ?@ ltxtquery | ltree | first array entry that matches ltxtquery; NULL if none | X


lib/rom/sql/extensions/postgres/types/ltree.rb
```diff
@@ -0,0 +1,140 @@
...
+            Sequel::SQL::PlaceholderLiteralString.new(string, [expr, query])
+          end
+
+          def build_array_query(query, array_type = 'lquery')
+            case query
+            when Sequel::Postgres::PGArray
+              if query.array_type == array_type
+                query
+              else
+                query.array_type = array_type
```
flash-gordon: We should never mutate incoming data, this goes against POLS

lib/rom/sql/extensions/postgres/types/ltree.rb
```diff
@@ -0,0 +1,140 @@
...
+          end
+
+          def match_any(type, expr, query)
+            array = build_array_query(query)
+            Attribute[SQL::Types::Bool].meta(sql_expr: custom_sql_expr(MATCH_ANY, expr, array))
+          end
+
+          private
+
+          def custom_sql_expr(string, expr, query)
```
flash-gordon: If this is used only for operators then having a more specific name like ‘operator_expr’ would be a better choice

GustavoCaso:
@flash-gordon I apply the feedback provided.
And added some documentation

solnic:
Should we close #244 in favor of this one?

solnic:
Thanks @GustavoCaso this is fantastic 🎉

PR #249, opened 2017-10-05
[PG] added meta data for geometric types

GustavoCaso:
@flash-gordon as we talk yesterday I updated the geometric Types.

### DIFF
--- a/lib/rom/sql/extensions/postgres/types/geometric.rb
+++ b/lib/rom/sql/extensions/postgres/types/geometric.rb
@@ -32,100 +32,114 @@ module ROM
         # The list of geometric data types supported by PostgreSQL
         # @see https://www.postgresql.org/docs/current/static/datatype-geometric.html
 
-        Point = SQL::Types.define(Values::Point) do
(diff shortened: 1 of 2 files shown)


lib/rom/sql/extensions/postgres/types/geometric.rb
```diff
@@ -32,100 +32,114 @@ module Types
         # The list of geometric data types supported by PostgreSQL
         # @see https://www.postgresql.org/docs/current/static/datatype-geometric.html
 
-        Point = SQL::Types.define(Values::Point) do
-          input do |point|
-            "(#{ point.x },#{ point.y })"
-          end
+        Point = Type('point') do
+            SQL::Types.define(Values::Point) do
```
flash-gordon: Looks like there is two extra spaces at the beginning of the line

solnic:
Please add specs too, I'm not sure what this PR actually adds.

GustavoCaso:
@solnic sorry for the late response I added the specs

PR #253, opened 2017-10-21
Extract common tree path object

solnic:
This is on top #252

### DIFF
--- a/lib/rom/sql/extensions/postgres/types/array_types.rb
+++ b/lib/rom/sql/extensions/postgres/types/array_types.rb
@@ -19,13 +19,11 @@ module ROM
             @base_write_type = base_write_type
             @base_read_type = base_read_type
             @constructor = proc { |db_type, member|
-              -> arr {
-                if member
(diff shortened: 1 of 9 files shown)


lib/rom/sql/extensions/postgres/types/array_types.rb
```diff
@@ -0,0 +1,72 @@
...
+
+          attr_reader :base_read_type
+
+          def initialize(base_write_type, base_read_type)
+            @elements = {}
+            @base_write_type = base_write_type
+            @base_read_type = base_read_type
+            @constructor = proc { |db_type, member|
+              -> arr {
+                if member
```
flash-gordon: this should be
```ruby
if member
  -> arr {  }
else
  -> arr {  }
end
```
i.e. the condition can be checked only once per type rather than once per cast

solnic: @flash-gordon ohh nice catch! I'll address this.

PR #255, opened 2017-10-23
Fix inferring for custom types primary keys

v-kolesnikov:


solnic:
Nice one, thank you :)

PR #263, opened 2017-11-10
Refactor instrumentation plugin and enable logging all sql

solnic:
This changes instrumentation plugin to be an sql-specific instrumentation. Unfortunately we can't re-use instrumentation from core, simply because in order to log all queries we are forced to hook into `log_connection_yield` which is defined on sequel's database objects (aka db connections). In the context of this method, we do not have access to relations, and I don't think it's possible to provide that, since it's a database object that's shared across all relations.


solnic:
I merged this in despite CI failure with jruby, because it fails on master too, so it's unrelated to changes made in this PR

PR #264, opened 2017-11-11
Move command timestamp plugin to rom-core

GustavoCaso:
This PR is part to fix https://github.com/rom-rb/rom/issues/467

This PR is part of one in `rom-core` https://github.com/rom-rb/rom/pull/469


cflipse:
I might suggest leaving the specs in place, so we can verify that the version included in core behaves the same ...

GustavoCaso:
Nice @cflipse. 

I did actually move all the test from `rom-sql` to `rom-core` https://github.com/rom-rb/rom/blob/b7db7febfb1aeb3c46f5a41cdecd1508373e4de9/core/spec/unit/rom/plugins/command/timestamps_spec.rb

cflipse:
Oh, I figured you _did_ move the tests along with the code; I wasn't suggesting that you hadn't.

Rather, leave the specs in here _as well_ until after the next release, so they can act as a canary, to let us know if anything gets out of sync between releases.  After release, or maybe a minor release, they can be removed.

GustavoCaso:
Thanks for the explanation @cflipse I will bring back the specs 😄

PR #267, opened 2017-11-16
Dockerize for easier development

Krule:
I wanted to propose a feature to `sql-rom` via PR but could not bundle the gem as I was missing dependencies. 

As I prefer not to keep dependencies I am not actively using on my host machine and it's my presumption a lot of other developers these days are doing the same, I would like to offer this small PR. It should add an easy way to start developing within docker container with all dependencies already present.

How to use? Well, you type `make dev` and, assuming you do have docker installed, that's it. You are in the container with dependencies available and can start working.


solnic:
This should be helpful indeed. Could you also add a couple sentences to the README about this?

Krule:
@solnic Updated

Krule:
Is this still considered useful?

GustavoCaso:
Maybe not make it by default to use docker, so depending on if you are using docker or not you should have a default when setting the `spec_helper` https://github.com/rom-rb/rom-sql/pull/267/files#diff-93830fa29d616f7c87903d08b5b1b29aR42

spec/spec_helper.rb
```diff
@@ -38,15 +39,15 @@
 if defined? JRUBY_VERSION
   DB_URIS = {
     sqlite: 'jdbc:sqlite:',
-    postgres: 'jdbc:postgresql://localhost/rom_sql',
-    mysql: 'jdbc:mysql://localhost/rom_sql?user=root&sql_mode=STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION&useSSL=false',
+    postgres: ENV.fetch('POSTGRES_DSN', 'jdbc:postgresql://localhost/rom_sql'),
+    mysql: ENV.fetch('MYSQL_DSN', 'jdbc:mysql://localhost/rom_sql?user=root&sql_mode=STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION&useSSL=false'),
```
Krule: @GustavoCaso That is exactly what is happening here. Use environment variable, and if not present, which is the default case, use "default" DSN.

GustavoCaso: Awesome, sorry I was not very familiar with `fetch` and not using block

GustavoCaso:
I do not see why not to merge it. I really think is a valuable addition

solnic:
@flash-gordon I'm OK with merging this in, you?

flash-gordon:
yeah, whatever works, I didn't test it :)

solnic:
One thing I'd change here, is `Makefile`, it's uncommon in the ruby land to use `make`, so could we change it to use rake tasks instead?

Krule:
I would limit `Rake` for `Ruby` things, but this is docker container management. Prefer to keep it separate and make is present almost everywhere.

timriley:
Let's use Rake tasks for this. It's a ruby project already, Rake is already in use, and Rake is perfectly suited to running external commands.

elskwid:
@Krule - we use a lot of Docker and Docker Compose on our projects and we opt to let the developers alias those commands as they see fit. We document them in the README.

This is especially helpful on platforms where they may not already have `ruby` or `make` or whatever. Then the set up is: "Install Docker and Docker Compose and run these commands".

splattael:
@Krule I've just tested the docker setup with a clean repo and it seems that `Gemfile.lock` to has to be present.

(Unfold for more details)

<details>

```Building gem
Step 1/7 : FROM ruby:2.4.2
 ---> 2a867526d472
Step 2/7 : RUN mkdir /gem
 ---> Using cache
 ---> 113e6877ebb3
Step 3/7 : WORKDIR /gem
 ---> Using cache
 ---> a3ed606352ca
Step 4/7 : COPY Gemfile Gemfile.lock rom-sql.gemspec ./
ERROR: Service 'gem' failed to build: COPY failed: stat /var/lib/docker/tmp/docker-builder021717139/Gemfile.lock: no such file or directory
Makefile:2: recipe for target 'build' failed
make: *** [build] Error 1
```

</details>

Not copying `Gemfile.lock` during `docker build` would help:

```diff
diff --git a/Dockerfile b/Dockerfile
index 22dec3e..6bdd15f 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM ruby:2.4.2
 RUN mkdir /gem
 WORKDIR /gem
 
-COPY Gemfile Gemfile.lock rom-sql.gemspec ./
+COPY Gemfile rom-sql.gemspec ./
 COPY lib/rom/sql/version.rb ./lib/rom/sql/
 
 RUN bundle install --jobs 8 --retry 5
```

WDYT?

Krule:
@splattael I think dependencies should be locked in order to have reproducible environment. But, as that is not the case, yes, `Gemfile.lock` is not supposed to be used as I did.

ianks:
Yes this is useful. It will help new contributors be productive more quickly. Also, I think we could port this to rom-factory as well.

solnic:
Doesn't work for me:

```
∞ [ruby-2.5.1][node-v8.12.0] rom-sql git(dockerize_for_easier_development) docker-compose build gem
Building gem
Step 1/7 : FROM ruby:2.4.2
 ---> 2a867526d472
Step 2/7 : RUN mkdir /gem
 ---> Using cache
 ---> 6707b99fdef9
Step 3/7 : WORKDIR /gem
 ---> Using cache
 ---> f5ad6d6e9666
Step 4/7 : COPY Gemfile rom-sql.gemspec ./
 ---> Using cache
 ---> 26e6de61585d
Step 5/7 : COPY lib/rom/sql/version.rb ./lib/rom/sql/
 ---> Using cache
 ---> aa9431de4c3f
Step 6/7 : RUN bundle install --jobs 8 --retry 5
 ---> Using cache
 ---> 9a6796c5e6b7
Step 7/7 : ADD . /gem/
 ---> b4e0ed529e56
Successfully built b4e0ed529e56
Successfully tagged rom-sql_gem:latest
∞ [ruby-2.5.1][node-v8.12.0] rom-sql git(dockerize_for_easier_development) docker-compose run --rm gem 'rspec'
Starting rom-sql_mariadb_1  ... done
Starting rom-sql_postgres_1 ... done
Could not find concurrent-ruby-1.0.5 in any of the sources
Run `bundle install` to install missing gems.
```

PR #271, opened 2017-12-15
Support for functions with any return type

### LINKED ISSUE #270: Support for functions with `Any` return type
Since we have `Types::Any` now, we can add support for defining functions w/o return type, in which case `Types::Any` will be set by default, ie:

``` ruby
select { fun(:count, :id).as(:count), concat(:first_name, :last_name).as(:full_name) }
```

Refs rom-rb/rom-rb.org#227

GustavoCaso:
Fixes #270 

Thanks to @solnic and @flash-gordon for the indications to implement this one.

At the end, we decided to go with `select { f(:count, :id).as(:count), concat(:first_name, :last_name).as(:full_name) }` for brevety

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## v2.3.1 to-be-released
+
+### Added
+
+* Support for functions with `Any` return type (GustavoCaso)
+
 ## v2.3.0 2017-11-17
 
 ### Added
--- a/lib/rom/sql/projection_dsl.rb
+++ b/lib/rom/sql/projection_dsl.rb
@@ -28,6 +28,10 @@ module ROM
         super || type(name)
       end
 
+      def f(name, attr)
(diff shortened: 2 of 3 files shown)


abrthel:
I wonder if naming it `function` and aliasing it to `f` would be a better idea? While I love the brevity of `f` it's not immediately apparent what it does and I know my first thought when trying to use a function would be to call well `function`. So it hurts discover-ability a bit.

GustavoCaso:
@abrthel @solnic I did change it to be call `function`and aliased to `f`

GustavoCaso:
@solnic @flash-gordon any update on this one?

flash-gordon:
@GustavoCaso pls add docs/changelog entry for this, then merge, thanks!

GustavoCaso:
@flash-gordon will do 👍

PR #274, opened 2018-01-24
Add Relation#as_hash

### LINKED ISSUE #272: Add Relation#as_hash
Sequel has a sweet feature where you can ask for a hash by providing an attribute name. We could expose it too, with a small improvement that it doesn't require an argument, and we default to the pk attribute so:

``` ruby
pages.select(:id, :title).to_hash
=> {{1=>{:id=>1, :title=>"Page 1"}, 2=>{:id=>2, :title=>"Page 2"}}

pages.select(:id, :title).to_hash(:title)
=> {{"Page 1"=>{:id=>1, :title=>"Page 1"}, "Page 2"=>{:id=>2, :title=>"Page 2"}}
```

GustavoCaso:
Resolves #272 

Add `Relation#as_hash`

@flash-gordon my first iteration was not delegating to `dataset`

```ruby
def as_hash(attribute = primary_key)
   to_a.each_with_object({}) do |tuple, result|
      result[tuple[attribute]] = tuple
   end
end
```

But since is already implemented in `Sequel` I thought it might be a good idea.

### DIFF
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -919,6 +919,24 @@ module ROM
           end
         end
 
+        # Returns hash with all tuples being
+        # the key of each the provided attribute
+        #
+        # @example default use primary_key
+        #   users.as_hash
+        #   # {1 => {id: 1, name: 'Jane'}}
+        #
+        # @example using other attribute
+        #   users.as_hash(:name)
+        #   # {'Jane' => {id: 1, name: 'Jane'}}
+        #
+        # @return [Hash]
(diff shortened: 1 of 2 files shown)


lib/rom/sql/relation/reading.rb
```diff
@@ -919,6 +919,25 @@ def each_batch(size: 1000)
           end
         end
 
+
```
solnic: looks like an extra empty line

solnic:
minor style issue to fix, other than that, please merge when addressed :)

PR #282, opened 2018-04-03
Add function case

wmaciejak:


lib/rom/sql/function.rb
```diff
@@ -132,6 +132,22 @@ def cast(expr, db_type = TypeSerializer[:default].call(type))
         Attribute[type].meta(sql_expr: ::Sequel.cast(expr, db_type))
       end
 
+
+      # Add a CASE clause for handling if/then logic
+      #
+      # @example
+      #   users.select { string::case({ "on" => "active" }, "inactive", :status).as(:activated) }
```
flash-gordon: `:status` should be `status`, i.e. be a method call rather than a symbol. Symbols work too but we don't advertise them since it's Sequel's API we're not aiming to support

flash-gordon: _I think_ status should be the first argument, it looks more inline with SQL, easier to remember.
In other words, there are two possible calls:
```ruby
case(matching, default_case)
case(expr, matching, default_case)
```

However, `default_case` is optional so we have to rely on a `is_a?(Hash)` check to distinguish these two calls
```ruby
case(matching, default_case)
case(expr, matching)
```

I know it's more work then simply wrapping a ready Sequel's API but we since have a chance to come up with a better design we should try.

spec/unit/function_spec.rb
```diff
@@ -61,6 +61,22 @@
...
+      it 'returns an sql expression' do
+        expect(func.case({ 1 => "first" }, "last", users[:id]).sql_literal(ds))
+          .to eql(%((CASE "users"."id" WHEN 1 THEN 'first' ELSE 'last' END)))
+      end
+    end
+
+    context 'when additional expression is not provided' do
+      it 'returns an sql expression' do
+        expect(func.case({ 1 => "first" }, "last").sql_literal(ds))
+          .to eql(%((CASE WHEN 1 THEN 'first' ELSE 'last' END)))
```
flash-gordon: This SQL is not valid, at least not in PostgreSQL/Oracle, it's better to have a working example, we don't want to confuse people :)

wmaciejak: Right, good point! I thought about literal value for `true`, but I'd forgotten that MySQL and PostgreSQL have a different mapping for Integer and string in that case.

lib/rom/sql/function.rb
wmaciejak: I noticed that case have one more use case. The matching argument might be a Hash or Array. 

Okay, anyway I modified implementation a little bit, but to be honest - Idea indeed is very good, but the implementation... I don't feel well with the quality of it. I'm always trying to omit and replace fetching args directly from the position in array e.g `args[2]` etc. ;)
Do you have a similar idea in your mind or I'm don't get something?

lib/rom/sql/function.rb
```diff
@@ -132,6 +132,27 @@ def cast(expr, db_type = TypeSerializer[:default].call(type))
         Attribute[type].meta(sql_expr: ::Sequel.cast(expr, db_type))
       end
 
+
```
v-kolesnikov: Seems like an extra blank line.

flash-gordon:
Sorry for the delay, I'll get back to it as soon as I finish a big chunk of work at my day job. I hope I'm close...

lib/rom/sql/function.rb
```diff
@@ -132,6 +132,26 @@ def cast(expr, db_type = TypeSerializer[:default].call(type))
...
+      # @return [ROM::SQL::Attribute]
+      #
+      # @api public
+      def case(*args)
+        if args.first.is_a?(Hash) || args.first.is_a?(Array)
+          Attribute[type].meta(sql_expr: ::Sequel.case(args[0], args[1]))
+        else
+          Attribute[type].meta(sql_expr: ::Sequel.case(args[1], args[2], args[0]))
+        end
+      end
```
solnic: Seems like changing this method's sig to `def case(*args, **opts)` should simplify things, no?

wmaciejak: Thanks for review @solnic !
Very good point, I wanted to omit `if else` statement, but I cannot achieve that because of sequel implementation of CaseExpression are pretty not friendly for wrappers ;)

This looking good for you?
```
      def case(*args, **opts)
        if opts[:expr]
          Attribute[type].meta(sql_expr: ::Sequel.case(*args, opts[:expr]))
        else
          Attribute[type].meta(sql_expr: ::Sequel.case(*args))
        end
      end
```

solnic: Yeah, that's what I meant :) I knew we can't avoid the conditional.

flash-gordon:
I merged this into a separate feature branch, I'll be working on this today since I finally got some spare time. I think I'll extend/change the API a bit for the case with `:expr`, it makes sense to add `#case` to the `Attribute` class instead.

PR #284, opened 2018-04-06
Optimize `pluck` to select only the column specified in SQL

joelvh:
This optimizes SELECT queries to only select the column needed for `pluck`


flash-gordon:
thanks!

Issue #279, opened 2018-03-21
has_many through join generates incorrect query


blelump:
Hi!

I'd like to report a problem with simple `has_many through` usage. Lets consider simple `users` has many `employees` case with intersection table `employees_users`. For a join on `employees`, ROM generates query:
```
SELECT `users`.`id`, `users`.`login` FROM `users` 
INNER JOIN `employees_users` ON (`users`.`id` = `employees_users`.`user_id`) 
INNER JOIN `employees` ON (`users`.`id` = `employees_users`.`user_id`) 
WHERE (`users`.`id` = 1) ORDER BY `users`.`id`
```

whereas it should be (compare join on `employees`):
```
SELECT `users`.`id`, `users`.`login` FROM `users` 
INNER JOIN `employees_users` ON (`users`.`id` = `employees_users`.`user_id`) 
INNER JOIN `employees` ON (`employees`.`id` = `employees_users`.`employee_id`) 
WHERE (`users`.`id` = 1) ORDER BY `users`.`id`
```

It helps a bit, when you'll provide `foreign_key` on relation associations, ie:
```
  conf.relation(:users) do
    schema(:users, infer: true) do
      associations do
        has_many :employees_users, foreign_key: :user_id
        has_many :employees, through: :employees_users, foreign_key: :employee_id
      end
    end
  end
```

but still, the join on `employees` in incorrect.

See the code below to recreate the problem.
```
rom = ROM.container(:sql, 'sqlite::memory') do |conf|
  conf.gateways[:default].use_logger(Logger.new($stdout))

  conf.default.create_table(:users) do
    primary_key :id
    column :login, String, null: false
  end

  conf.default.create_table(:employees) do
    primary_key :id
    column :name, String, null: false
  end

  conf.default.create_table(:employees_users) do
    foreign_key :employee_id, :employees
    foreign_key :user_id, :users
  end

  conf.relation(:users) do
    schema(:users, infer: true) do
      associations do
        has_many :employees_users
        has_many :employees, through: :employees_users
      end
    end
  end

  conf.relation(:employees) do
    schema(:employees, infer: true) do
      associations do

      end
    end
  end

  conf.relation(:employees_users) do
    schema(:employees_users, infer: true) do
      associations do
        belongs_to :user
        belongs_to :employee
      end
    end
  end
end

class UserRepo < ROM::Repository[:users]
  def by_id(id)
    users.join(:employees)
      .by_pk(id)
      .one
  end
end

[...]

PR #289, opened 2018-05-22
Fix has_many through join query

doriantaylor:
It looks as if I stumbled across [the same issue](https://github.com/rom-rb/rom-sql/issues/279#issuecomment-391043656). This should fix it.

I have not included a unit test because I am not sure how to test it effectively. However, that the tests passed both before and after the change suggest the join conditions are not currently being verified.

### DIFF
--- a/spec/integration/associations/many_to_many_spec.rb
+++ b/spec/integration/associations/many_to_many_spec.rb
@@ -37,6 +37,38 @@ RSpec.describe ROM::SQL::Associations::ManyToMany, helpers: true do
         specify { expect(assoc.combine_keys).to eql(id: :task_id) }
       end
 
+      describe '#dataset' do
+        it 'actually performs the correct join' do
+          # get the resulting dataset from an ordinary join
+          ds = tasks.join(:tags).dataset
+
+          # now get the joins
+          joins = ds.opts[:join]
+
+          # joins should be an array of two elements
+          expect(joins).to be_a Array
+          expect(joins.length).to be 2
+
+          # collect a list of join pairs
+          join_coll = []
+          joins.each do |j|
+            # if this isn't true then this loop will crash
+            expect(j).to be_a Sequel::SQL::JoinOnClause
+
+            join_coll << j.on.args.map(&:to_sym)
+          end
+
+          # here is what the join column pairs should look like
+          join_ok = [
+            [:tasks__id, :task_tags__task_id],
+            [:task_tags__tag_id, :tags__id],
+          ].freeze
+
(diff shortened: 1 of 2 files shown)


doriantaylor:
Yeah I have no idea how to test this; it should test the contents of `container.relations[:a].join(:b).dataset.opts[:join]` but I have no idea how to get at that in the test suite. It looks like they aren't the same class of objects in there.

doriantaylor:
OK I figured out how to test it. Unit test added.

solnic:
Thanks for the fix. To test it properly just add another example in [this spec](https://github.com/rom-rb/rom-sql/blob/master/spec/unit/relation/inner_join_spec.rb#L119-L123). It looks like we have a similar setup there with `:tags` association. The specific example I linked to uses `inner_join(tags)` so I suppose we need to also check `inner_join(:tags)`, or maybe this test is passing by an accident due to the way test data is setup.

spec/integration/associations/many_to_many_spec.rb
```diff
@@ -37,6 +37,38 @@
...
+          join_ok = [
+            [:tasks__id, :task_tags__task_id],
+            [:task_tags__tag_id, :tags__id],
+          ].freeze
+
+          # mkay?
+          expect(join_coll).to eql join_ok
+        end
+      end
+
```
solnic: Please remove this and see my other comment in this PR.

doriantaylor:
From what I could tell, the conditions being tested by the many-to-many test would have passed irrespective of the join conditions. That's why I elected to verify the contents of the resulting `Sequel::Dataset` objects.

To clarify, are you asking me to *move* that test to the inner join spec or rewrite it entirely?

spec/integration/associations/many_to_many_spec.rb
doriantaylor: Noted, [see mine](https://github.com/rom-rb/rom-sql/pull/289#issuecomment-391360071).

solnic:
> To clarify, are you asking me to move that test to the inner join spec or rewrite it entirely?

It should be a different type of test. I believe it's a matter of using the same strategy as other examples in inner_join specs, but fixtures must be improved so that we don't get correct results by an accident.

solnic:
@ianks we just need a good unit test which uses the right fixtures that can verify expected behavior

doriantaylor:
Feel free to take a crack at it; I'm kinda on another planet right now…

solnic:
Closing in favor of #328

PR #290, opened 2018-05-27
Add Relation#query

flash-gordon:
This adds support for subqueries in the select part. It will probably work in `where` as well but I need to test it. I'm also not certain about naming.

### DIFF
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -937,6 +937,18 @@ module ROM
           dataset.as_hash(attribute)
         end
 
+        # Turn a relation into a subquery. Can be used
+        # for selecting a column with a subquery or
+        # restricting the result set with a IN (SELECT ...) condtion.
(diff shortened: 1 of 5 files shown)


lib/rom/sql/relation/reading.rb
```diff
@@ -937,6 +937,12 @@ def as_hash(attribute = primary_key)
           dataset.as_hash(attribute)
         end
 
+        def query
```
flash-gordon: @solnic it's three lines of code

solnic:
Wouldn't it work automagically if we had `Relation#to_sql_literal`?

flash-gordon:
@solnic we want it to act as an attribute actually, as an example, support for arithmetic operations: https://github.com/flash-gordon/ledger/blob/e17cbba09576210f2b3aff4bc3743a66bbed14c0/lib/ledger/persistence/relations/accounts.rb#L24-L41

PR #292, opened 2018-06-27
reject attributes that do not have dataset when qualifying

### LINKED ISSUE #288: rom-sql tries to qualify expressions without dataset
from https://github.com/hanami/model/issues/475

```ruby
rel.select_append { `'hello'`.as(:bar) }.qualified
```
ends up with `NoMethodError: undefined method 'dataset' for nil:NilClass. I think we shouldn't touch attributes without dataset, they're expressions, not table columns.

GustavoCaso:
Fixes #288 

@solnic @flash-gordon hope this helps with the issue.

Any comments I'll be happy to address them.

### DIFF
--- a/lib/rom/sql/attribute.rb
+++ b/lib/rom/sql/attribute.rb
@@ -62,6 +62,7 @@ module ROM
       # @api public
       def qualified(table_alias = nil)
         return self if qualified? && table_alias.nil?
+        return meta(qualified: false) unless qualifiable?
 
         case sql_expr
         when Sequel::SQL::AliasedExpression, Sequel::SQL::Identifier, Sequel::SQL::QualifiedIdentifier
@@ -112,6 +113,15 @@ module ROM
       # @api public
       def qualified?
         meta[:qualified].equal?(true) || meta[:qualified].is_a?(Symbol)
+      end
+
+      # Return if an attribute is qualifiable
+      #
+      # @return [Boolean]
+      #
+      # @api public
+      def qualifiable?
+        !source.nil?
       end
 
       # Return a new attribute marked as a FK
--- a/spec/unit/relation/qualified_spec.rb
+++ b/spec/unit/relation/qualified_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe ROM::Relation, '#qualified' do
     it 'qualifies all attributes' do
       qualified = relation.qualified
 
-      expect(qualified.schema.all?(&:qualified)).to be(true)
+      expect(qualified.schema.all?(&:qualified?)).to be(true)
     end
 
     it 'qualifies virtual attributes' do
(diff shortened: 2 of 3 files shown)


lib/rom/sql/attribute.rb
```diff
@@ -62,6 +62,7 @@ def canonical
       # @api public
       def qualified(table_alias = nil)
         return self if qualified? && table_alias.nil?
+        return meta(qualified: false) unless source
```
solnic: Let's add:

``` ruby
def qualifiable?
  ! source.nil?
end
```

and do:

``` ruby
return meta(qualified: false) unless qualifiable?
```

This way it's more explicit, otherwise we're adding special meaning to `source` that is not clearly visible without reading `qualified` method and understanding what it's doing.

spec/unit/relation/qualified_spec.rb
```diff
@@ -17,9 +17,21 @@
                     qualified.
                     group(:id)
 
-      expect(qualified.schema.all?(&:qualified)).to be(true)
+      expect(qualified.schema.all?(&:qualified?)).to be(true)
 
       expect(qualified.to_a).to eql([{ id: 1, count: 1 }, { id: 2, count: 1 }])
     end
+
+    it 'does not qualifies attribute without dataset' do
```
solnic: `does not qualifies` => `does not qualify`

spec/unit/relation/qualified_spec.rb
```diff
@@ -17,9 +17,21 @@
...
+      qualified = relation.select_append { `'hello'`.as(:bar) }.qualified
+
+      expect(qualified.schema.all?(&:qualified?)).to be(false)
+    end
+
+    it 'does not raise NoMethodError when qualifies attribute without dataset' do
+      expect {
+        relation.select_append { `'hello'`.as(:bar) }.qualified
+      }.to_not raise_error
+    end
```
solnic: I wouldn't write a test that checks if it raises or not, a better way is to simply check expected behavior, ie:

``` ruby
qualified = relation.select_append { `'hello'`.as(:bar) }.qualified

expect(qualified.schema[:bar]).to_not be(nil) # or some better check?
```

GustavoCaso: I guess by removing this spec and simply leaving the one on above should work right? https://github.com/rom-rb/rom-sql/pull/292/files#diff-f159d65edb9c79561a91c5f0b3c6f4e3R25

GustavoCaso:
@solnic Thank you so much for the comments, as always great advice ❤️

spec/unit/relation/qualified_spec.rb
solnic: @GustavoCaso yeah sounds good

solnic:
Please rebase/squash + merge. We should also cherry-pick this into `release-2.x` branch and push a new bug fix release. I can deal with that.

PR #293, opened 2018-06-28
update Dry::Types::Int to Dry::Types::Integer and ROM::SQL::Types::Int to ROM::SQL::Types::Integer

GustavoCaso:
@flash-gordon hope this help 😄


solnic:
Is this a breaking change? 😅

GustavoCaso:
@solnic I hope not is just an internal change. 😄

flash-gordon:
rom-sql in master depends on dry-types 0.13. This means master broke compatibility with 2.x in any case

solnic:
Ahh ok so I guess this can be just merged in

PR #294, opened 2018-06-28
reject attributes that do not have dataset when qualifying

GustavoCaso:
@solnic here is the cherry pick for the `release-2.5`

### DIFF
--- a/lib/rom/sql/function.rb
+++ b/lib/rom/sql/function.rb
@@ -60,6 +60,13 @@ module ROM
         )
       end
 
+      # @see Attribute#qualified?
+      #
+      # @api private
+      def qualified?(table_alias = nil)
+        meta[:func].args.all?(&:qualified?)
+      end
+
       # @see ROM::SQL::Attribute#is
       #
       # @api public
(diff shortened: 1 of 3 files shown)


lib/rom/sql/function.rb
```diff
@@ -60,6 +60,13 @@ def qualified(table_alias = nil)
         )
       end
 
+      # @see Attribute#qualified?
+      #
+      # @api private
+      def qualified?(table_alias = nil)
+        meta[:func].args.all?(&:qualified?)
```
ianks: @GustavoCaso is `meta[:func]` guaranteed to always be defined here? I think we should be defensive here and check that it exists.

spec/unit/relation/qualified_spec.rb
```diff
@@ -17,9 +17,15 @@
                     qualified.
                     group(:id)
 
-      expect(qualified.schema.all?(&:qualified)).to be(true)
+      expect(qualified.schema.all?(&:qualified?)).to be(true)
 
       expect(qualified.to_a).to eql([{ id: 1, count: 1 }, { id: 2, count: 1 }])
     end
+
+    it 'does not qualify attribute without dataset' do
```
ianks: What do you mean by "without a dataset"?

lib/rom/sql/function.rb
solnic: Good point. It's not guaranteed, because a new function object is treated as a function builder, whatever method you call on it, will result in setting `meta[:func]`.

solnic:
Sorry but looks like we didn't manage to handle a new patch release. This will be in 3.0.0 instead.

PR #303, opened 2018-08-18
Add a simple DSL for joining relations

flash-gordon:
It shouldn't even break existing code

### DIFF
--- a/lib/rom/sql/join_dsl.rb
+++ b/lib/rom/sql/join_dsl.rb
@@ -1,0 +1,8 @@
+require 'rom/sql/restriction_dsl'
+
+module ROM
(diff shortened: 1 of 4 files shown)


abrthel:
This is awesome! I literally ran into this problem yesterday and resorted to manual sql. THANK YOU

v-kolesnikov:
I have a couple of dangerous places in my code where I had join relations with Sequel DSL and worried about correct tables names. Hope it would be more easier now with native ROM DSL. Thank you @flash-gordon 🎉

flash-gordon:
I thought about looking for cases yet to cover but, as far as I can tell, this is good to go.

PR #304, opened 2018-08-23
Add SQL::Relation#unfiltered

flash-gordon:


solnic:
Man, I wish `#all` wasn't a method known from other libs, because it would make much more sense. Anyway, I'm OK with `unfiltered`.

flash-gordon:
@solnic yeah, but I don't think it'll be used very often

Issue #310, opened 2018-11-20
Simplify setting up migrator


solnic:
Currently gateway options does not support passing migrator options, so if you need to customize it, you need to manually instantiate a connection, then migrator, then pass it as an option to rom configuration, which is silly.

This is what we could have instead:

``` ruby
config = ROM::Configuration.new(:sql, "postgresql://localhost/rom", migrator: { path: "my_migrations_path" })
```

This is a matter of tweaking logic in [migration extension](https://github.com/rom-rb/rom-sql/blob/master/lib/rom/sql/migration.rb#L100-L104) because right now it only supports `migrator` object in options.

cllns:
Looks like this can be closed, it was addressed in #311

PR #311, opened 2018-11-21
Simplify setting up migrator

rawburt:
https://github.com/rom-rb/rom-sql/issues/310

### DIFF
--- a/lib/rom/sql/migration.rb
+++ b/lib/rom/sql/migration.rb
@@ -98,7 +98,7 @@ module ROM
 
       # @api private
       def initialize(uri, options = EMPTY_HASH)
-        @migrator = options.fetch(:migrator) { Migrator.new(connection) }
+        @migrator = create_migrator(options[:migrator])
 
         self.class.instance ||= self
       end
@@ -145,6 +145,21 @@ module ROM
         end
 
         migrator.auto_migrate!(self, schemas, options)
+      end
+
+      private
+
+      # Create a `Migrator`. If `migrator_option` is a `Hash`, use it as options to `Migrator.new`.
+      #
+      # @api private
+      def create_migrator(migrator_option)
+        return Migrator.new(connection) unless migrator_option
+
+        if migrator_option.is_a?(Hash)
+          Migrator.new(connection, migrator_option)
+        else
+          migrator_option
+        end
       end
     end
   end
--- a/spec/integration/gateway_spec.rb
+++ b/spec/integration/gateway_spec.rb
@@ -62,6 +62,23 @@ RSpec.describe ROM::SQL::Gateway, :postgres do
         container.gateways[:default].run_migrations
       end
     end
+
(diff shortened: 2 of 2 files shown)


v-kolesnikov:
So, how I can build a configuration without any migrator as I could before by passing `migrator: nil` to `ROM::Configuration.new` ? It was useful.

solnic:
@v-kolesnikov oh I haven't thought about it, we should tweak the code to make it possible again

vizvamitra:
Hi. I came across this PR while solving the following problem: 

Several months ago we've created a migration with an extra digit in it's timestamp. Not a big deal for prod, so we haven't noticed it for 4 months. But recently I added another migration that changes same stuff, and `rake db:drop db:create db:migrate` started failing. Investigation showed that ROM applies migrations with extra digit in the timestamp after all the rest. 

I now try to rename them, but it is not straightforward how: `db:migrate` crashes when some of the migration files are missing, and `ROM::Configuration.new(..., migrator: {allow_missing_migration_files: true})` doesn't seem to have any effect on `db:migrate` task. 

So, I wonder: is it something that was missed by this PR, or is it by intent?

cllns:
@vizvamitra Please open a new issue for that. 

I'd drop a `binding.irb` into the code changed here to see if/where that option is getting dropped. 

But please note that `allow_missing_migration_files` is ["very risky to use"](https://sequel.jeremyevans.net/rdoc/classes/Sequel/Migrator.html). And we [advise against renaming migrations](https://sequel.jeremyevans.net/rdoc/files/doc/migration_rdoc.html#label-Modifying+existing+migrations). Is there some other way you can fix the problem, perhaps with a new migration?

vizvamitra:
Thank you very much for your response and the links. Not sure if it is a bug, since you discourage usage of this option, so I've created a topic in discussions board instead.

> Is there some other way you can fix the problem

We'll deal with this, I already have a plan that'll help not to break anything. No worries and thank you for your care. I only wanted to highlight the issue.

> where that option is getting dropped

Basically it happens because:

- `ROM::SQL::Migrations::Migrator` lacks `option :allow_missing_migration_files, default: -> { false }` line, so the option is being filtered out in the initializer
- And since it doesn't accept any other options for `Sequel::Migrator`, doesn't mix anything into `options` passed to `#run`

Something like this could work, I guess:

```ruby
module ROM
  module SQL
    module Migration
      class Migrator
        extend Initializer

        # ...

        option :allow_missing_migration_files, default: -> { false }

        def run(options = {})
          options = migrator_options.merge(options)
          Sequel::Migrator.run(connection, path.to_s, options)
        end

        private

        def migrator_options
          {
            allow_missing_migration_files:,
            # maybe something else, candidates: :table, :column, :use_advisory_lock
            # See https://sequel.jeremyevans.net/rdoc/classes/Sequel/Migrator.html#method-method-c-run
          }
        end
      end
    end
  end
end
```

PR #312, opened 2018-11-22
Allow to install pry via Bundler under Truffleruby

deepj:
Without this, the most specs fail on `cannot load such file -- pry` under Truffleruby.

This first baby step heading to Truffleruby support. Currently there are still some problems under Truffleruby.

But this PR can bring one issue. Bundler can recognize `truffleruby` since 1.16.5 and later. Unfortunately only for specs thought. See https://github.com/bundler/bundler/blob/v1.16.5/lib/bundler/current_ruby.rb

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ end
 
 group :test do
   gem 'pry-byebug', platforms: :mri
-  gem 'pry', platforms: %i(jruby rbx)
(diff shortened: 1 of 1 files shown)


solnic:
@deepj I created `truffleruby` branch where we can keep things up to date. Commit from this branch is there.

deepj:
@solnic ok, thx!

Issue #316, opened 2019-01-12
Not possible to order by aliased attribute


waiting-for-dev:
**Describe the bug**

Trying to call `#order` in a relation with an attribute which is aliased in the schema generates an invalid SQL query, adding `AS 'alias_name'` in the `ORDER BY` clause.

**To Reproduce**

```ruby
require "rom"
require "rom-sql"
require "dry/types"

module Types
  include Dry::Types.module
end

rom = ROM.container(:sql, 'sqlite::memory') do |conf|
  conf.default.create_table(:users) do
    primary_key :id

    column :username, String, null: false
  end

  conf.relation(:users) do
    schema(infer: true) do
      attribute :username, Types::String.meta(alias: :name)
    end
  end
end

users = rom.relations[:users]

order = users.order(users[:username])
puts order.dataset.sql
# => SELECT `users`.`username` AS 'name', `users`.`id` FROM `users` ORDER BY `users`.`username` AS 'name'
order.to_a
# => #<Sequel::DatabaseError: SQLite3::SQLException: near "AS": syntax error>
```

**Expected behavior**

Order should work with aliased attributes.

**Your environment**

- Affects my production application: **YES**
- Ruby version: 2.5

**Resources**
- [Original discussion](https://discourse.rom-rb.org/t/default-datasets-joins-and-aliases/299).

PR #318, opened 2019-01-16
[WIP] Fix aliased attributes sql literals

waiting-for-dev:
References #316

This is still a WIP to fix referenced bug: tests are missing and some
previous have been broken. A more comprehensive solution needs to be
found, so I ask for any feedback or help.

When sequel creates a sql expression from an object, it tries to call
one the two following methods in this order:

- `sql_literal_append`
- `sql_literal`

rom-sql is implementing `sql_literal`, which only takes as argument the
sequel datastore. However, `sql_literal_append` takes two arguments: the
datastore and the sql string built so far.

The partial solution implemented in this commit matches the sql string
at the moment of invocation in order to detect whether we are to the
left of the `FROM table` clause or to the right of it. If we are to the
left we are referencing a column from the `SELECT` clause, so we keep
appending `name AS alias`. However, if we are to the right, what we need
is the canonical sql literal representation (i.e. `WHERE name = 1`).

This is obviously very hacky. I haven't found any way to inspect the
dataset or any other component in order to have a solid knowledge about
which sql clause is being constructed. So I'm afraid it could be the
only possible fix.

In case something in the same vein of this solution is to be
implemented, we should think the most comprehensive way to match the sql
query built so far. For example, the current solution would not work for
an `UPDATE table SET name as alias = 1...` statement. We should also
study whether it could break edge corner cases, like a column name
containing `FROM table` in its name.

### DIFF
--- a/lib/rom/sql/attribute.rb
+++ b/lib/rom/sql/attribute.rb
@@ -275,20 +275,13 @@ module ROM
         Function.new(type).concat(self, sep, other)
       end
 
-      # Sequel calls this method to coerce an attribute into SQL string.
-      #
-      # If the attribute is aliased and it is on the right of the
-      # `FROM table` keyword, we need to use its canonical name.
+      # Sequel calls this method to coerce an attribute into SQL string
       #
       # @param [Sequel::Dataset] ds
       #
       # @api private
-      def sql_literal_append(ds, sql)
-        if aliased? && sql.match?(/ FROM ["`]#{source.dataset}["`] /)
-          canonical.sql_literal_append(ds, sql)
-        else
-          ds.literal_append(sql, sql_expr)
-        end
+      def sql_literal(ds)
+        ds.literal(sql_expr)
       end
 
       # Sequel column representation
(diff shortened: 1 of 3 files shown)


solnic:
What about handling this at relation/schema level? This would be more explicit and less risky. We can convert attributes depending on the context in which they are used (this is how it works already in the relation API).

waiting-for-dev:
Thanks for your feedback @solnic . I didn't realize it was possible at that level. Cool, I'll dig into it and try to find something better.

waiting-for-dev:
Not sure if it is the same issue but, for a `:foo` aliased attribute the following doesn't work:

```ruby
relation.where { relation[:foo].like("something") }
```

The above fails with `#like` not being a method. Instead, one must do:

```ruby
relation.where { relation[:foo].canonical.like("something") }
```

For now, I just leave a note here so I don't forget.

waiting-for-dev:
What do you think of something like last commit, @solnic ? If it's ok, then we have to do something similar to `order` and any other clause where we need it.

solnic:
@waiting-for-dev could you re-push this branch to rom-sql repo and we could take it from there?

waiting-for-dev:
Ok. Closing in favour of #325 I cleaned the commit history there.

PR #321, opened 2019-02-25
Properly support multiple arguments passed to #pluck

timriley:


lib/rom/sql/relation/reading.rb
```diff
@@ -148,15 +148,19 @@ def map(key = nil, &block)
...
+        #   users.pluck(:id, :name)
+        #   # [[1, "Jane"] [2, "Joe"]]
         #
         # @return [Array]
         #
         # @api public
-        def pluck(name)
-          select(name).map(name)
+        def pluck(*names)
+          select(*names).map(names.length == 1 ? names.first : names)
```
timriley: I'm not super happy with the ternary here, but without it we end up with a result like `[[1], [2]]` when plucking a single value only. An alternative would be to `flatten(1)` the result in the case of a single name being passed, but that seems even worse. Any thoughts?

solnic: This is fine, it can always be improved later on if somebody comes up with a better idea.

PR #324, opened 2019-04-18
Update to work with decoupled alias options

solnic:
Refs rom-rb/rom#512

PR #326, opened 2019-04-18
Fix mapper compiler reading attribute alias

waiting-for-dev:
After merging #324 to master, `rom` test suite (once `rom-sql`
dependency was updated to the new master ref) was failing due to a change
introduced by error on reading the attribute alias from the mapper
compiler.

### DIFF
--- a/lib/rom/sql/mapper_compiler.rb
+++ b/lib/rom/sql/mapper_compiler.rb
@@ -4,10 +4,10 @@ module ROM
   module SQL
     class MapperCompiler < ROM::MapperCompiler
       def visit_attribute(node)
-        name, _, meta = node
+        name, _, meta_options = node
 
-        if meta[:wrapped]
-          [name, from: self.alias]
+        if meta_options[:wrapped]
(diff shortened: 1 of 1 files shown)


flash-gordon:
what are the reasons for these renamings?

waiting-for-dev:
The last item in the AST now contains both meta and options:

https://github.com/rom-rb/rom/blob/master/core/lib/rom/attribute.rb#L396

I see this naming as temporal (but explicit), because I think that everything that now is in `meta` should end up as an `option`, in order to be able to rely on type inference while still being able to set those options. Just like we can do right now with `alias:`.

flash-gordon:
Thanks for the explanation, I agree with the sentiment, we'll see if we can pull all data out of meta, it's not obvious because using types as intermediate objects that pass stuff around can be useful too. We'll probably keep data in both places.

solnic:
I'm merging this despite broken CI because there multiple tweaks that are needed to make all this stuff work again, it's gonna be easier to push to master directly.

PR #330, opened 2019-05-02
Add ROM::SQL::Relation#nullify

ianks:
I noticed there wasn't a simple way to nullify a relation. This functionality is common in Sequel and Rails (`ActiveRecord::Relation#none`). Will make it easier to short circuit any DB interactions when necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,21 @@
+## Unreleased
+
+### Added
+
+- Support for `.nullify` on a relation for easier usage of the null-object
+  pattern (@ianks)
+  ```ruby
+  users.nullify.count # => will always be 0
+  users.nullify.to_a # => will always be empty ([])
+  ```
+
+[Compare v3.0.1...master](https://github.com/rom-rb/rom-sql/compare/v3.0.0...master)
+
 ## 3.0.1 2019-05-05
 
 ### Fixed
 
-* Using joins with a relation object as the target and a block-based conditions produces correct query now (issue #331) (solnic)
+- Using joins with a relation object as the target and a block-based conditions produces correct query now (issue #331) (solnic)
 
 [Compare v3.0.0...v3.0.1](https://github.com/rom-rb/rom-sql/compare/v3.0.0...v3.0.1)
 
@@ -10,7 +23,7 @@
 
 ### Added
 
-* Join DSL so that you can use arbitrary conditions when joining relations (flash-gordon)
+- Join DSL so that you can use arbitrary conditions when joining relations (flash-gordon)
   ```ruby
   users.join(tasks) { |users:, tasks:|
     tasks[:user_id].is(users[:id]) & users[:name].is('John')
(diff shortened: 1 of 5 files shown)


solnic:
Can you come up with a usage example?

ianks:
@solnic I made a PR to rom-rb.org docs which has an example. One use case I encountered recently is when we were mapping an API response with IDs, when the response is from the API is empty we know that the resulting relation will be empty, so we can avoid a database call. A common example would be a "guest" user who will never have access to certain resources, so you can bypass all DB interaction for these resources using `nullify`.

ianks:
Fixed typos and merge conflicts.

solnic:
I have mixed feelings about this but I won't be blocking this addition. WDYT @flash-gordon ?

ianks:
Can you expand on your mixed feelings a bit?

solnic:
@ianks when I want to avoid unnecessary db interactions I just don't use objects that talk to the db, rather than "nullifying" them, but again, this is just my limited experience. I've simply never needed a feature like this (or at least that's what I think :))

ianks:
Makes sense. It's definitely a great tool to have in the toolbox when it's needed. That being said, it's not an "everyday" type of tool. Hopefully you can trust me on that 😀

flash-gordon:
In my opinion, this is a perfect example of cutting corners. Problem is, from the rom-sql POV we only need to describe an abstract SQL statement. Adding `#nullify` is orthogonal to this DSL thus requires extending it with an option or a thing equivalent to it. As a proof, Sequel relies on `Object#extend` to support this feature which is a quite uncommon technique in Ruby (I guess mostly because of historical performance issues but probably also because of making debugging process more difficult). The whole [support for `nullify`](https://github.com/jeremyevans/sequel/blob/035b214b9600ed623e38d4215bca711f09cc7d72/lib/sequel/extensions/null_dataset.rb) looks rather ad hoc, I cannot be certain it will work for cases beyond simple SQL statements. I would rather use `.where('1 = 0')` or something similar because it lies within the existing DSL, the cost of one extra DB roundtrip is negligible compared to possible debugging complications caused by `.nullify` applied somewhere in the code. My guess is only a handful of people will be aware this method even exists.

That said, some people compare libraries apples-to-apples disregarding differences in approaches, so I think we can add support for `nullify` for this very reason.

ianks:
That is one perspective @flash-gordon. I tend to think that `where('1 = 0')` requires an equal amount of trust in the implementation and internal knowledge of that library that someone couldn't override that by running `.or(some_other_condition: true)`. I happen to know that this is not part of the rom-sql API but many wouldnt. `nullify` clearly states the intention of the API.

ianks:
Wrote that before I saw the merge, thank you for seeing the use case 😀

flash-gordon:
There's no unscoped `or` in the API since there's no use case for this. Anyway, calling `dataset.sql` returns you the SQL to be run so that you can see all conditions compiled whereas detecting nullified datasets is barely possible.

solnic:
WDYT about turning this into a plugin?

flash-gordon:
Agreed. 
Interesting how AR makes it more explicit a relation was nullified https://github.com/rails/rails/blob/99df469ddc3dc4be611f4a17f76d14af9ab01bbd/activerecord/lib/active_record/relation/query_methods.rb#L794
It also uses `Object#extend` btw, though they could use composition instead.

solnic:
Cool. This should be a plugin IMO because it's not what everybody needs.

PR #333, opened 2019-05-15
Wrap JSON primitives before passing them to Sequel (fix #327)

### LINKED ISSUE #327: Data coercion when writing to postgres JSONB column not working
**Describe the bug**

I am getting

```
ROM::SQL::DatabaseError (PG::DatatypeMismatch: ERROR:  column "value" is of type jsonb but expression is of type integer)
```

when attempting to write to a postgres JSONB column using both `Relation#insert` and `Relation#command(:create)`.

**To Reproduce**
1. Create a table called `attributes` in your db in postgres with a `name::VARCHAR` column and a `value::JSONB` column.
2. Set up ROM::Relation as such:
```
class AttributesRelation < ROM::Relation[:sql]
  gateway :default

  schema(:attributes, infer: true)
end
```
3. Attempt both of the following:
```
attributes = ROM.env.relations[:attributes]
attributes.insert(name: "valid name", value: 1)
```
```
attributes = ROM.env.relations[:attributes]
attributes.command(:create).call(name:"valid name", value: 1)
```
**Expected behavior**

I expected the integer value to automatically be coerced into valid JSON before attempting to write to the database. Effectively:
```
attributes.insert(name: "valid name", value: 1.to_json)
```
```
attributes = ROM.env.relations[:attributes]
attributes.command(:create).call(name: "valid name", value: 1.to_json)
```
(these both work btw)

**Your environment**

- Affects my production application: **NO**
- Ruby version: 2.6.2
- Rails version: 5.2.3
- OS: MacOS Mojave Version 10.14.3

**Note**
Please let me know if coercing data into the JSON format before inserting said data is not an intended feature. I have worked around this by creating a changeset which coerces the data in the `value` key into valid JSON before attempting insert.

This works:
```
class NewAttributeChangeset < ROM::Changeset::Create
  map do
    map_value :value, ->(v) { v.to_json }
  end
end
```
```
attributes.changeset(NewAttributeChangeset, { name: "valid name", value: 1 }).commit
```

Any advice you have on best practices around this is much appreciated. Thank you for taking the time to investigate.

flash-gordon:
This adds support for writing primitive values to JSON(B) columns. Sequel has recently added support for it. On the rom-sql's side we only need to change the wrapper.

Other values are interpreted as JSONOp as they did before. We could revise this behavior in the next major version, it was added for compatibility with Hanami < 1.0 but I'm not sure if we still need it.

### DIFF
--- a/lib/rom/sql/extensions/postgres/types/json.rb
+++ b/lib/rom/sql/extensions/postgres/types/json.rb
@@ -16,16 +16,45 @@ module ROM
           end
         end
 
-        JSON = Type('json') do
-          (SQL::Types::Array | SQL::Types::Hash)
-            .constructor(Sequel.method(:pg_json))
-            .meta(read: JSONRead)
-        end
+        if Sequel.respond_to?(:pg_json_wrap)
+          json_types = [
+            SQL::Types::Array,
+            SQL::Types::Hash,
+            SQL::Types::Integer,
+            SQL::Types::Float,
+            SQL::Types::String,
+            SQL::Types::Nil,
+            SQL::Types::True,
+            SQL::Types::False
+          ]
 
-        JSONB = Type('jsonb') do
-          (SQL::Types::Array | SQL::Types::Hash)
-            .constructor(Sequel.method(:pg_jsonb))
-            .meta(read: JSONRead)
+          JSON = Type('json') do
+            json_cast_methods = Hash.new(:pg_json)
+            json_cast_methods.update(json_types.each_with_object({}) { |t, h| h[t.primitive] = :pg_json_wrap })
+
+            json_types
+              .reduce(:|)
+              .constructor { |value| Sequel.public_send(json_cast_methods[value.class], value) }
+              .meta(read: JSONRead)
+          end
+
+          JSONB = Type('jsonb') do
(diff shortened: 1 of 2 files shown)


lib/rom/sql/extensions/postgres/types/json.rb
```diff
@@ -16,16 +16,45 @@ module Types
...
+            SQL::Types::False
+          ]
+
+          JSON = Type('json') do
+            json_cast_methods = Hash.new(:pg_json)
+            json_cast_methods.update(json_types.each_with_object({}) { |t, h| h[t.primitive] = :pg_json_wrap })
+
+            json_types
+              .reduce(:|)
+              .constructor { |value| Sequel.send(json_cast_methods[value.class], value) }
```
solnic: IIRC `__send__` is more appropriate (if you wanted to use priv methods), otherwise it should be `public_send`.

Gemfile
```diff
@@ -4,6 +4,8 @@ git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }
 
 gemspec
 
+gem 'sequel', github: 'jeremyevans/sequel'
```
solnic: I wonder if we should set up a matrix with sequel from master + latest released version?

PR #334, opened 2019-05-15
Properly qualify relations when in Relation#union

ianks:
When Sequel performs a union of two datasets, it will alias the resulting union as `t1` by default. As a result, `select_append` anything that required a qualified column will not work after performing a union. 

To fix this, this PR ensures we qualify the schema when performing a union. If relation_a and relation_b are of the `Relation::Name`, then we use that name as the alias. If they are different, we default the alias to `#{relation_a.name}__#{relation_b.name}`. This is not _technically_ required, but IMO makes the relation more easily inspectable/debuggable, and is more resistant to cases when user's manually qualify columns.

### DIFF
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -807,7 +807,16 @@ module ROM
         #
         # @api public
         def union(relation, options = EMPTY_HASH, &block)
-          new(dataset.__send__(__method__, relation.dataset, options, &block))
+          # We use the original relation name here if both relations have the
+          # same name. This makes it so if the user at some point references
+          # the relation directly by name later on things won't break in
+          # confusing ways.
+          same_relation = name == relation.name
+          alias_name =  same_relation ? name : "#{name.to_sym}__#{relation.name.to_sym}"
+          opts = { alias: alias_name.to_sym, **options }
+
+          new_schema = schema.qualified(opts[:alias])
+          new_schema.(new(dataset.__send__(__method__, relation.dataset, opts, &block)))
         end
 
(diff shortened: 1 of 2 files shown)


lib/rom/sql/relation/reading.rb
```diff
@@ -807,7 +807,16 @@ def select_group(*args, &block)
...
+          # We use the original relation name here if both relations have the
+          # same name. This makes it so if the user at some point references
+          # the relation directly by name later on things won't break in
+          # confusing ways.
+          same_relation = name == relation.name
+          alias_name =  same_relation ? name : "#{name.to_sym}__#{relation.name.to_sym}"
+          opts = { alias: alias_name.to_sym, **options }
+
+          new_schema = schema.qualified(opts[:alias])
+          new_schema.(new(dataset.__send__(__method__, relation.dataset, opts, &block)))
```
flash-gordon:
```ruby
new(dataset.__send__(__method__, relation.dataset, opts, &block), schema: new_schema)
```
would be better here

spec/unit/relation/union_spec.rb
```diff
@@ -3,17 +3,76 @@
...
+      end
+    end
+
+    context 'when the relations unioned have different names' do
+      let(:relation1) { relation.where(id: 1).select(:id, :name) }
+      let(:relation2) { tasks }
+
+      it 'qualifies the table as the concatenated relation names' do
+        result = relation1.union(relation2)
+        expect_to_have_qualified_name(result, :users__tasks)
```
flash-gordon: I'm not sure if we even need to test this but anyway

flash-gordon:
Good catch

PR #335, opened 2019-05-23
Honor association views when wrapping

ianks:
Previously, when wrapping a relation, the view for the association would not be honored. This would make the output of `.wrap` different from combine, and cause unexpected behavior.

This PR addresses that by first applying the view to the association target if there is one. 

Also, since `.wrap` with prefix attributes, any previously aliased attributes would fail loudly since `Sequel::SQL::AliasedExpression` does not respond to `.as`. To fix this, we now will handle alias `Sequel::SQL::AliasedExpression` by constructing a new instance with the new alias name.

### DIFF
--- a/lib/rom/sql/associations/core.rb
+++ b/lib/rom/sql/associations/core.rb
@@ -14,6 +14,14 @@ module ROM
 
           target.where(target_key => target_pks)
         end
+
+        # @api private
+        def wrapped
+          new_target = view ? target.send(view) : target
+          to_wrap = self.class.allocate
+          to_wrap.send(:initialize, definition, options.merge(target: new_target))
+          to_wrap.wrap
+        end
       end
     end
   end
--- a/lib/rom/sql/attribute.rb
+++ b/lib/rom/sql/attribute.rb
@@ -5,6 +5,8 @@ require 'rom/attribute'
 
 require 'rom/sql/type_extensions'
 require 'rom/sql/projection_dsl'
+require 'rom/sql/attribute_wrapping'
+require 'rom/sql/attribute_aliasing'
 
 module ROM
   module SQL
@@ -12,6 +14,9 @@ module ROM
     #
     # @api public
     class Attribute < ROM::Attribute
+      include AttributeWrapping
+      include AttributeAliasing
+
       OPERATORS = %i[>= <= > <].freeze
       NONSTANDARD_EQUALITY_VALUES = [true, false, nil].freeze
       META_KEYS = %i(index foreign_key target sql_expr qualified).freeze
@@ -25,21 +30,6 @@ module ROM
       def self.[](*args)
         fetch_or_store(args) { new(*args) }
       end
-
-      # Return a new attribute with an alias
-      #
-      # @example
-      #   users[:id].aliased(:user_id)
-      #
-      # @return [SQL::Attribute]
-      #
-      # @api public
-      def aliased(alias_name)
-        super.with(name: name || alias_name).meta(
-          sql_expr: sql_expr.as(alias_name)
-        )
-      end
-      alias_method :as, :aliased
 
       # Return a new attribute in its canonical form
       #
--- a/lib/rom/sql/attribute_aliasing.rb
+++ b/lib/rom/sql/attribute_aliasing.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module ROM
+  module SQL
+    # @api private
+    module AttributeAliasing
+      # Return a new attribute with an alias
+      #
+      # @example
+      #   users[:id].aliased(:user_id)
+      #
+      # @return [SQL::Attribute]
+      #
+      # @api public
+      def aliased(alias_name)
+        new_name, new_alias_name = extract_alias_names(alias_name)
+
+        super(new_alias_name).with(name: new_name).meta(
+          sql_expr: alias_sql_expr(sql_expr, new_alias_name)
+        )
+      end
+      alias as aliased
+
+      private
+
+      # @api private
+      def alias_sql_expr(sql_expr, new_alias)
+        case sql_expr
+        when Sequel::SQL::AliasedExpression
+          Sequel::SQL::AliasedExpression.new(sql_expr.expression, new_alias, sql_expr.columns)
+        else
+          sql_expr.as(new_alias)
+        end
+      end
+
+      # @api private
+      def extract_alias_names(alias_name)
+        new_name, new_alias_name = nil
+
+        if wrapped? && aliased?
+          # If the attribute is wrapped *and* aliased, make sure that we name the
+          # attribute in a way that will map the the requested alias name.
+          # Without this, the attribute will silently ignore the requested alias
+          # name and default to the pre-existing name.
+          new_name = "#{meta[:wrapped]}_#{options[:alias]}".to_sym
+
+          # Essentially, this makes it so "wrapped" attributes aren't true
+          # aliases, in that we actually alias the wrapped attribute, we use
+          # the old alias.
+          new_alias_name = options[:alias]
+        else
+          new_name = name || alias_name
+          new_alias_name = alias_name
+        end
+
+        [new_name, new_alias_name]
+      end
+    end
+  end
+end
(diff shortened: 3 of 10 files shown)


ianks:
Can i get a review on this?

lib/rom/sql/relation/reading.rb
```diff
@@ -1025,6 +1025,31 @@ def nullify
...
+          others = names.map do |name|
+            association = associations[name]
+            opts = association.options
+            definition = association.definition
+            target = association.target
+            new_target = association.view ? target.send(association.view) : target
+            to_wrap = association.class.allocate
+            to_wrap.send(:initialize, definition, opts.merge(target: new_target))
+            to_wrap.wrap
+          end
```
solnic: This indicates that there's a missing abstraction. It seems like having a new method on association objects that will handle all of this (and provide a meaningful name at the same time) would be a good idea. Could you come up with something?

spec/integration/wrap_spec.rb
```diff
@@ -86,6 +86,39 @@
...
         end
       end
+
+      context 'using association with a view' do
+        before do
+          conf.relation(:users) do
+            auto_map false
+
+            schema(infer: true) do
+            end
```
solnic: No need for this empty block

spec/integration/wrap_spec.rb
```diff
@@ -86,6 +86,39 @@
...
+          end
+        end
+
+        it 'includes the extra attributes' do
+          result = tasks.wrap(:users).to_a
+
+          result.each do |task|
+            expect(task[:users]).to eql(extra_attribute: 'testing')
+          end
```
solnic: It would be better to materialize this relation and compare it with the expected resulting structure. If, for whatever reason, `result` is empty, that `each` won't do a thing and the spec will continue to pass.

spec/unit/attribute_spec.rb
```diff
@@ -94,6 +94,13 @@
     end
   end
 
+  describe '#aliased' do
+    it 'can alias an previously aliased attribute' do
+      expect(users[:id].as(:uid).as(:uuid).sql_literal(ds)).
+        to eql(%("users"."id" AS "uuid"))
+    end
+  end
```
solnic: It would be good to extract this into a separate PR. I think it's a good fix that can be shipped separately.

lib/rom/sql/attribute.rb
```diff
@@ -36,7 +36,7 @@ def self.[](*args)
       # @api public
       def aliased(alias_name)
         super.with(name: name || alias_name).meta(
-          sql_expr: sql_expr.as(alias_name)
+          sql_expr: alias_sql_expr(sql_expr, alias_name)
```
flash-gordon: it looks like we should have our own composable wrapper for aliased expressions, that case analysis in `alias_sql_expr` is leaking

ianks:
@flash-gordon I noticed a weird error when wrapping an aliased function. essentially, the attribute name has an extra prefix that is never unwrapped, so for the failing spec https://github.com/rom-rb/rom-sql/pull/335/commits/27f46040ef5a21552dbc1ccd760991b543019b44, the wrapped key is `users_extra_attribute` rather than `extra_attribute`. Do you have any clues as to why this might happen? Stuck on it right now, but would like to get it fixed :)

ianks:
Fixed the issue with aliasing functions. But now I've discovered a new regression in 5.0, aliasing attributes in wrapped relations does not work. So working on that now.

lib/rom/sql/attribute.rb
ianks: I think this will be challenging to do in this PR, i'm not sure exactly how to approach it since from what i can see, `sql_expr` is an actual Sequel::* class everywhere so it may have unexpected breakage. Can we postpone this refactoring?

ianks:
Alright, so I *believe* this PR is ready. It might not be the prettiest, but at least the behavior is correct now. With this patch, our (large) production app can now pass all specs on rom 5!

### What this patch addresses

- views are now honored in wrapped associations
- attribute aliasing on wrapped attributes used the correct alias name
- proper aliasing for functions with wrapped attributes

Excited to get this in so we do not have to run off a fork, and contribute upstream to bugfixes, features, etc.

- Ian

lib/rom/sql/relation/reading.rb
```diff
@@ -1011,6 +1011,35 @@ def unfiltered
...
+        # @see http://sequel.jeremyevans.net/rdoc-plugins/files/lib/sequel/extensions/null_dataset_rb.html
+        # @example result will always be empty, regardless if records exists
+        #   users.where(name: 'Alice').nullify
+        #
+        # @return [SQL::Relation]
+        #
+        # @api public
+        def nullify
+          new(dataset.__send__(__method__))
+        end
```
solnic: This seems to be a left-over from another PR, right? 😄

lib/rom/sql/associations/core.rb
```diff
@@ -14,6 +14,14 @@ def preload(target, loaded)
...
           target.where(target_key => target_pks)
         end
+
+        # @api private
+        def wrapped
+          new_target = view ? target.send(view) : target
+          to_wrap = self.class.allocate
+          to_wrap.send(:initialize, definition, options.merge(target: new_target))
+          to_wrap.wrap
+        end
```
solnic: Could you remind me why you had to use `Object#allocate` here?

spec/unit/attribute_spec.rb
```diff
@@ -94,6 +94,13 @@
     end
   end
 
+  describe '#aliased' do
+    it 'can alias an previously aliased attribute' do
```
psparrow: nit: "an" => "a"

lib/rom/sql/attribute_aliasing.rb
```diff
@@ -0,0 +1,61 @@
...
+      # @example
+      #   users[:id].aliased(:user_id)
+      #
+      # @return [SQL::Attribute]
+      #
+      # @api public
+      def aliased(alias_name)
+        new_name, new_alias_name = extract_alias_names(alias_name)
+
+        # super.with(name: name || alias_name).meta(
```
psparrow: cut commented code

lib/rom/sql/associations/core.rb
ianks: `Association#new` is overridden with defaults that do not work for this use case

ianks:
PR comments have been addressed :)

lib/rom/sql/associations/core.rb
solnic: @ianks could we tweak the underlying implementation in a way that would not require using `allocate` + `send(:initialize, ...)` here? something like `new(target: view ? target.send(view) : target)` should work.

solnic: ...sorry, I meant `with(target: view ? target.send(view).target)`, that's the conventional method we use in rom.

ianks: does this require changes to rom-core? @solnic

solnic: @ianks OK I see what's going on, it's due to the way `Associations::Core.new` is implemented. This is unfortunate, I'll do something about it later so for now let's leave this `allocate` trick here.

solnic:
@flash-gordon could you take a look as well, please?

lib/rom/sql/mapper_compiler.rb
```diff
@@ -7,11 +7,20 @@ def visit_attribute(node)
...
         else
           [name]
         end
       end
+
+      private
+
+      def extract_wrapped_name(node)
+        _, _, meta_options = node
+        unwrapped_name = meta_options[:alias].to_s
```
flash-gordon: in ruby 2.7, `Symbol#to_s` start returning frozen strings, this code will break

ianks:
@flash-gordon @solnic fixed the last issue. code climate seems to be failing for an unrelated reason. merge? :smile:

flash-gordon:
@ianks nice, please squash the commits, I'll merge it shortly after

ianks:
@flash-gordon All squashed.

PR #365, opened 2020-02-12
Remove ORDER to make COUNT faster

joelvh:
Lifted from `graphql-ruby`: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/pagination/sequel_dataset_connection.rb#L19

### DIFF
--- a/lib/rom/sql/plugin/pagination.rb
+++ b/lib/rom/sql/plugin/pagination.rb
@@ -62,7 +62,7 @@ module ROM
           #
           # @api public
           def total
-            dataset.unlimited.count
+            dataset.unlimited.order(nil).count
           end
 
           # Return total number of pages


flash-gordon:
I genuinely wonder which DBMS doesn't optimize it itself...

solnic:
@flash-gordon are you saying we should make this optimization on a per-db type basis?

flash-gordon:
@solnic I'm saying that no DB doesn't do it by itself. Actually, using `count` with `order by` won't even work in PG, which means Sequel already ignores the ordering. This PR, however, reveals a bug in `total` because it uses `unlimited` datasets, which is plain wrong. If you want to paginate over a filtered relation, you'll get incorrect results. I'm amazed this hasn't been reported. I'll file an issue.

joelvh:
@flash-gordon nice catch #366

Issue #370, opened 2020-03-03
Aliased attributes cause various relation functions to output malformed sql


abrthel:
**Describe the bug**

When functions like `where`, `join`, `order` on relation are given aliased attribute instances they
produce malformed sql. This can happen with the method signature api and block level DSL api.

For example:
`user_relation.order(user_relation[:name])` would output -> `ORDER BY "users"."name" AS 'user_name'`

**To Reproduce**
The following gist reproduces most some of the issues I've found with aliases 

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rom', '5.2.1'
  gem 'rom-sql, '3.2.0'
  gem 'sqlite3'
  gem 'rspec'

  # gem 'pry'
  # gem 'pry-byebug'
end

RSpec.configure do |config|
  config.disable_monkey_patching!
  config.warnings = true
  config.filter_run_when_matching :focus
end

RSpec.describe "With aliased attributes return correct results when" do

  let(:users) { rom.relations[:users] }
  let(:tasks) { rom.relations[:tasks] }

  let(:rom) do
    rom = ROM.container(:sql, 'sqlite::memory') do |conf|
      conf.default.create_table(:users) do
        primary_key :id
        column :name, String, null: false
      end

      conf.default.create_table(:tasks) do
        primary_key :id
        column :name, String, null: false
        foreign_key :user_id, :users, null: false
      end

      conf.relation(:users) do
        schema(infer: true) do
          attribute :name, alias: :user_name
        end
      end

      conf.relation(:tasks) do
        schema(:tasks, infer: true) do
          attribute :user_id, alias: :task_key

          associations do
            belongs_to(:user)
          end
        end
      end
    end
  end

  before do
    rom.relations[:users].insert(id: 1, name: 'Jane')
    rom.relations[:users].insert(id: 2, name: 'John')
    rom.relations[:users].insert(id: 3, name: 'Shelly')

    rom.relations[:tasks].insert(id: 1, name: 'Pick up milk', user_id: 2)
    rom.relations[:tasks].insert(id: 2, name: 'Call Insurance', user_id: 3)
    rom.relations[:tasks].insert(id: 3, name: 'Cancel Phone Plan', user_id: 3)
  end

  describe '#order' do
    let(:result) { [{id: 1, user_name: "Jane"}, {id: 2, user_name: "John"}, {id: 3, user_name: "Shelly"}] }

    it 'given block dsl' do
[...]

PR #371, opened 2020-03-03
Fix malformed sql when sql methods given aliased attribute instances

abrthel:
Fixes rom-rb/rom-sql#370

The main issue came from `Sequel::AliasedExpressions` being set as the attributes `:sql_expr` on every aliased attribute during schema finalization. It was a consequence of calling `:qualified` on all of the attributes.

As a consequence, the fix was to force "canonical" attributes for the schemas canonical instance. This allowed much of the `schema.canonical` calls littered throughout the sql reader methods to just automatically work.

### DIFF
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -477,7 +477,7 @@ module ROM
           if block
             new(dataset.order(*args, *schema.canonical.order(&block)))
           else
-            new(dataset.__send__(__method__, *args, &block))
+            new(dataset.__send__(__method__, *canonicalize_args(args), &block))
           end
         end
 
@@ -1100,6 +1100,26 @@ module ROM
             end
           else
             raise ArgumentError, "+other+ must be either a symbol or a relation, #{other.class} given"
+          end
+        end
+
+        # Coerces {ROM::SQL::Attribute} arguments into their
+        # canonical form
+        #
+        # Used by relation functions that accept attributes but don't
+        # want aliased attributes to generate invalid sql
+        #
+        # @api private
+        def canonicalize_args(args)
+          if args.is_a?(Array)
+            args.map do |arg|
+              next arg unless arg.is_a?(ROM::SQL::Attribute)
+              arg.canonical
+            end
+          elsif args.is_a?(ROM::SQL::Attribute)
+            arg.canonical
+          else
+            arg
           end
         end
       end
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -1080,7 +1080,7 @@ module ROM
             elsif block
               __join__(type, other, JoinDSL.new(schema).(&block), opts)
             else
-              new(dataset.__send__(type, other.to_sym, join_cond, opts, &block))
+              new(dataset.__send__(type, other.to_sym, canonicalize_args(join_cond), opts, &block))
             end
           elsif other.is_a?(Sequel::SQL::AliasedExpression)
             new(dataset.__send__(type, other, join_cond, opts, &block))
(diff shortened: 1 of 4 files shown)


lib/rom/sql/relation/reading.rb
```diff
@@ -1102,6 +1102,32 @@ def __join__(type, other, join_cond = EMPTY_HASH, opts = EMPTY_HASH, &block)
...
+        # Coerces {ROM::SQL::Attribute} arguments into their
+        # canonical form
+        #
+        # Used by relation functions that accept attributes but don't
+        # want aliased attributes to generate invalid sql
+        #
+        # @api private
+        def canonicalize_args(args)
+          if args.is_a?(Array)
+            args.map {|arg|
```
solnic: Codacy Issue found: [Space between { and | missing.](https://app.codacy.com/gh/rom-rb/rom-sql/pullRequest?prid=5107465)

solnic:
Codacy Here is an overview of what got changed by this pull request:


```diff

Issues
======
- Added 4
           
```


See the complete overview on [Codacy](https://app.codacy.com/gh/rom-rb/rom-sql/pullRequest?prid=5107465&bid=16796141)

lib/rom/sql/relation/reading.rb
```diff
@@ -1102,6 +1102,32 @@ def __join__(type, other, join_cond = EMPTY_HASH, opts = EMPTY_HASH, &block)
...
+              next arg unless arg.is_a?(ROM::SQL::Attribute)
```
solnic: Codacy Issue found: [Add empty line after guard clause.](https://app.codacy.com/gh/rom-rb/rom-sql/pullRequest?prid=5107465)

lib/rom/sql/relation/reading.rb
```diff
@@ -1102,6 +1102,32 @@ def __join__(type, other, join_cond = EMPTY_HASH, opts = EMPTY_HASH, &block)
...
         end
+
+        # Coerces {ROM::SQL::Attribute} arguments into their
+        # canonical form
+        #
+        # Used by relation functions that accept attributes but don't
+        # want aliased attributes to generate invalid sql
+        #
+        # @api private
+        def canonicalize_args(args)
```
solnic: Codacy Issue found: [Perceived complexity for canonicalize_args is too high. [8/7]](https://app.codacy.com/gh/rom-rb/rom-sql/pullRequest?prid=5107465)

lib/rom/sql/relation/reading.rb
```diff
@@ -1102,6 +1102,32 @@ def __join__(type, other, join_cond = EMPTY_HASH, opts = EMPTY_HASH, &block)
...
+          if args.is_a?(Array)
+            args.map {|arg|
+              next arg unless arg.is_a?(ROM::SQL::Attribute)
+              arg.canonical
+            }
+          elsif args.is_a?(Hash)
+            args.each_with_object({}) {|(k, v), h|
```
solnic: Codacy Issue found: [Space between { and | missing.](https://app.codacy.com/gh/rom-rb/rom-sql/pullRequest?prid=5107465)

solnic:
I think this should be fixed in the schema. Aliases shouldn't be set by default in the sql expression because they are only needed when projecting a relation (so using `select` and similar).

abrthel:
I actually agree and investigated the option for awhile but kept coming up with solutions that I was pretty sure weren't going to be approved.

To put the change in the schema, we also have to separate Attribute#qualified from returning aliased expressions during finalization. No problems, until you do something like `users[:id].as(:key).qualified` now we need the attribute to build a qualified identifier that's aliased. So kind of contradicting concerns.

We could make `:alias` option just a config option then add another meta tag/option or look for AliasedExpression that tells `Attribute#aliased?` the attribute was actually aliases, but we would also need change `Attribute#aliased` to accept no parameters and then by default look for the alias config option.

In concept does that sound like something you'd accept?

solnic:
@abrthel seems like we're missing an abstraction here, which is "qualifying for projection" vs just qualifying. The former needs to include aliases (as `... AS ...` part in the SQL), the latter must not include it. The reason why we qualify schemas upon finalization is to end up with datasets that use fully qualified columns in `SELECT` part of the underlying query. We could change that to use something like `qualified_for_projection` and change `qualified` to not include aliases.

abrthel:
Makes sense, I'll fix-up this pull request either later today or tomorrow morning.

abrthel:
Close in favour of rom-rb/rom-sql#373

PR #388, opened 2020-12-17
Add `:pg_streaming` extension to support streaming tuples

ianks:
This PR wraps the `sequel_pg`'s streaming functionality up in a neat little bow. We have a maintain a patch that adds this and have been using it in prod for quite awhile. Figured it was time to release it to the world!

### DIFF
--- a/lib/rom/sql/gateway.rb
+++ b/lib/rom/sql/gateway.rb
@@ -81,6 +81,7 @@ module ROM
       def initialize(uri, options = EMPTY_HASH)
         @connection = connect(uri, options)
         load_extensions(Array(options[:extensions]))
+        Notifications.trigger("configuration.gateway.connected", connection: @connection)
 
         @options = options
 
@@ -247,4 +248,6 @@ module ROM
       end
     end
   end
+
+  Configuration.register_event("configuration.gateway.connected")
 end
--- a/spec/integration/plugins/pg_streaming_spec.rb
+++ b/spec/integration/plugins/pg_streaming_spec.rb
@@ -1,48 +1,46 @@
 # frozen_string_literal: true
 
 RSpec.describe "Plugins / :pg_streaming", seeds: true do
-  with_adapters do
-    if metadata[:postgres]
-      include_context "users and tasks"
+  with_adapters(:postgres) do
+    include_context "users and tasks"
 
-      before do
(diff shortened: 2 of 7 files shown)


lib/rom/sql/gateway.rb
```diff
@@ -247,4 +251,6 @@ def transaction_runner(_)
       end
     end
   end
+
+  Configuration.register_event("configuration.gateway.db.connected")
```
solnic, line 255: @ianks could we call this `"configuration.gateway.connected"` instead?

solnic:
@ianks please rebase, I've fixed issues with the spec suite

lib/rom/sql/gateway.rb
ianks: yes

ianks:
@solnic for whatever reason, i think the specs are still failing after i rebased with master.

also, should i remove the changelog entry here as well?

solnic:
> @solnic for whatever reason, i think the specs are still failing after i rebased with master.

so, this is bizarre but I can confirm that the specs fail in this particular branch. They are green on master on CI and on my local machine. I also tried different sqlite3 versions and I could reproduce the failures with both 1.3.x and 1.4.x.

This will be tough to debug, maybe the new configuration event causes it, or maybe conditional require of `sequel_pg`, it's really hard to say :/ 

> also, should i remove the changelog entry here as well?

yes please, I'll add an entry when merging

ianks:
@solnic So it seems you are right, requiring `sequel_pg` when other adapters is causing breaking of other specs. To remedy this, I made another CI matrix for `DB_ADAPTER`, so they are all tested in isolation. Let me know your thoughts!

ianks:
woot! think i got this in a solid place now. @solnic et al, lmk if you would like any more adjustments 💓

spec/integration/plugins/pg_streaming_spec.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+RSpec.describe "Plugins / :pg_streaming", seeds: true do
+  with_adapters do
```
solnic, line 4: I believe this could simply use `with_adapters(:postgres) do ... end` without the need for that extra `if`

PR #399, opened 2021-03-04
Add support for streaming composite relations

ianks:
This patch makes streaming for composite relations by applying the `right` side to each streamed tuple of the `left`.


solnic (approved):

solnic:
@ianks just pushed https://rubygems.org/gems/rom-sql/versions/3.3.3!

PR #401, opened 2021-03-29
Updating association docs to mention `combine`.

pat:
`assoc` is for querying association data from a relation, but `combine` is for loading associations within the context of the main relation.

Suggesting this because I feel the existing heading seemed to match what I was looking for, but the text/code underneath it didn't. Maybe I'm getting my terms/ideas mixed up though?


solnic:
Thanks man, this makes it more clear!

PR #406, opened 2022-03-14
Add migration_options to be able to use the default rake migration tasks with custom opts.

wuarmin:
Hey,
I want to use the default rake migration tasks, but need to customize the table opt. This PR would fix my need.

thanks and best regards


solnic:
Can you rebase this on top of release-3.5 and set it as base branch in this PR? Master is now unstable, targeting 4.0 release.

wuarmin:
Ok, thanks! I opened a new PR #407

PR #421, opened 2023-06-03
[Bug] Relation#union fails on virtual columns

dgollahon:
- When merging relations with virtual columns the qualified projection (added in #334) causes the virtual column to be moved into the outer select. This means that column will always be hardcoded to `NULL` instead of selecting from the result of the union. This is a [not-that-esoteric technique with unions](https://stackoverflow.com/a/39766372). This also worked in ROM prior to #334.

This PR is really an issue/bug report just in the form of a failing test. I don't have a good idea as to what the right solutin is but I figured it would be helpful to at least have a very concrete demonstration of the issue.

### DIFF
--- a/spec/unit/relation/union_spec.rb
+++ b/spec/unit/relation/union_spec.rb
@@ -72,6 +72,94 @@ RSpec.describe ROM::Relation, "#union" do
       end
     end
 
+    context "when the relations unioned have different names and virtual columns" do
+      let(:relation1) { relation.where(id: 1).select(:id, :name).select_append { [`NULL`.as(:title)] } }
+      let(:relation2) { tasks.select(:id).select_append { [`NULL`.as(:name)] }.select_append(:title) }
+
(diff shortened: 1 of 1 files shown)


spec/unit/relation/union_spec.rb
```diff
@@ -72,6 +72,94 @@
...
+    context "when the relations unioned have different names and virtual columns" do
+      let(:relation1) { relation.where(id: 1).select(:id, :name).select_append { [`NULL`.as(:title)] } }
+      let(:relation2) { tasks.select(:id).select_append { [`NULL`.as(:name)] }.select_append(:title) }
+
+      it "qualifies the table as the concatenated relation names" do
+        # this produces SQL like
+        # SELECT
+        #   `users__tasks`.`id`,
+        #   `users__tasks`.`name`,
+        #   NULL AS 'title'
```
dgollahon, line 84: this is the problematic column select. if this got transformed to `users__tasks.title`, all would be well

PR #426, opened 2024-06-27
Fixes Rails 7.2 dep warning

bendangelo:
Updated argument to color: true. This stops the rails 7.2 dep warning.


flash-gordon:
This should check for the rails version then

bendangelo:
Ok done.

lib/rom/sql/extensions/rails_log_subscriber.rb
```diff
@@ -15,20 +15,31 @@ def sql(event)
...
       end
+
+      private
+
+      def color_option
+        if Gem::Version.new(Rails::VERSION::STRING) >= Gem::Version.new('7.1')
+          { color: true }
+        else
+          true
+        end
```
flash-gordon, line 41: I have two suggestions

1. You can assign it to a constant `COLOR_OPTION = if ...` will work just as fine but with only 1 check rather than a check per every call
2. You use `Rails` constant but the plugin depends on active-support, you should check its version instead. For instance, I have a project where active-support is present but rails isn't.

flash-gordon (changes requested):

PR #427, opened 2024-06-27
Fixes Rails 7.2 dep warning

bendangelo:
Same fix at the last but for release 3.6


flash-gordon:
Thanks, I addressed it in main and 3.6.4 which is now out

PR #429, opened 2024-07-02
Support table aliases on associated relations

cllns:
I discovered this bug the other week. If you call `join` with a relation object, `table_alias` doesn't work at all. It only works with the name a symbol.

### DIFF
--- a/lib/rom/sql/relation/reading.rb
+++ b/lib/rom/sql/relation/reading.rb
@@ -1107,6 +1107,8 @@ module ROM
               end
 
               new(dataset.__send__(type, other.name.dataset.to_sym, join_cond, join_opts))
+            elsif opts[:table_alias]
+              new(dataset.__send__( type, other.name.key, join_cond, opts))
             else
               associations[other.name.key].join(type, self, other)
             end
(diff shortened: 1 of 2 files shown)


lib/rom/sql/relation/reading.rb
```diff
@@ -1086,13 +1086,16 @@ def coerce_conditions(conditions)
         # @api private
         def __join__(type, other, join_cond = EMPTY_HASH, opts = EMPTY_HASH, &block)
           if other.is_a?(Symbol) || other.is_a?(ROM::Relation::Name)
-            if join_cond.equal?(EMPTY_HASH) && !block
-              assoc = associations[other]
+            assoc = associations.key?(other) && associations[other]
+
+            if join_cond.eql?(EMPTY_HASH) && !block
```
cllns, line 1091:
```suggestion
            if (join_cond.nil? || join_cond.empty?) && !block
```
I'd rather do this. It'd allow `nil` or `{}` or `[]` as join_cond, which are all valid IMO.

I'm not sure why this was `.equal?`, since that does object equality (so even passing in `{}` would be false since it's a different though identical object).

Issue #423, opened 2023-11-05
Problem with :create command and custom types


katafrakt:
## Describe the bug

When using custom write type and read type to modify the data before inserting / after fetching data from the database, the `create` command passes original unmodified data to read type, sometimes causing exceptions.

## To Reproduce

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem "json"
  gem "rom-sql"
  gem "sqlite3"
end

require 'rom'
require 'rom-repository'

rom = ROM.container(:sql, 'sqlite::memory') do |conf|
  conf.default.create_table(:test) do
    primary_key :id
    column :value, String, null: false
  end

  class TestRelation < ROM::Relation[:sql]
    WriteType = Dry.Types.Constructor(String) { |value| JSON.dump({content: value}) }
    ReadType = Dry.Types.Constructor(String) { |value| JSON.parse(value)["content"] }

    schema(:test, infer: true) do
      attribute :value, WriteType, read: ReadType
    end
  end

  conf.register_relation(TestRelation)
end

class TestRepo < ROM::Repository[:test]
  commands :create
end

record = TestRepo.new(rom).create(value: "test value")
```

This causes:

```
/home/katafrakt/.asdf/installs/ruby/3.2.2/lib/ruby/3.2.0/json/common.rb:216:in `parse': unexpected token at 'test value' (JSON::ParserError)
	from /home/katafrakt/.asdf/installs/ruby/3.2.2/lib/ruby/3.2.0/json/common.rb:216:in `parse'
	from rom_custom_type_problem.rb:21:in `block in <class:TestRelation>'
	from /home/katafrakt/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/dry-types-1.7.1/lib/dry/types/constructor/function.rb:17:in `call'
	from /home/katafrakt/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/dry-types-1.7.1/lib/dry/types/constructor.rb:81:in `call_unsafe'
	from /home/katafrakt/.asdf/installs/ruby/3.2.2/lib/ruby/gems/3.2.0/gems/dry-types-1.7.1/lib/dry/types/schema/key.rb:46:in `call_unsafe'
[...]

PR #432, opened 2024-10-27
Fix inconsistent behaviour across adapters with read types and "create" command

katafrakt:
During the investigation for #423 I discovered that for SQLite and MySQL (or basically for everything aside from PostgreSQL) the read types are applied twice. This works well if the read type allows it, but if it does not, it fails for these adapters, making behaviour between adapter inconsistent.

The solution here is to use `relation.dataset.where` instead of `relation.where` in default implementation of `insert` method of `Create` command. `relation.where` already applies the read types on materialization, while `relation.dataset.where` does not. This is also consistent with how `Update` command behaves.

With that in place, we always call `finalize` on "raw" data from the database, avoiding double application of a read type.

The PR consists of two commits. First one just adds tests replicating the problem. The second one just changes the code.

### DIFF
--- a/lib/rom/sql/commands/create.rb
+++ b/lib/rom/sql/commands/create.rb
@@ -46,7 +46,7 @@ module ROM
         # @api private
         def insert(tuples)
           pks = tuples.map { |tuple| relation.insert(tuple) }
-          relation.where(relation.primary_key => pks).to_a
+          relation.dataset.where(relation.primary_key => pks).to_a
         end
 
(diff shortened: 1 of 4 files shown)


katafrakt:
These CI failures look weird. Did my changes cause them?

katafrakt:
@flash-gordon should fixes like this target `release-3.7` branches instead of `main`?

flash-gordon:
@katafrakt yes, please, change the base!

katafrakt:
@flash-gordon done

